PR #172, opened 2017-02-20
Some progress on migration

oskay:
Three changes made: 
signalUpdateProgress.emit(0, 0, 0) -> signalUpdateProgress.emit(0, 0)
tostring -> tobytes
QtGui.QMessageBox ->QtWidgets.QMessageBox  (many locations)

With these changes, I am able to launch within qt5 and (to the extent that I have tested) knit successfully.  (This is tested on a Mac, running Python 2.7.13.)

While this does represent some progress, pop-up dialog boxes are still not working; I will open a separate issue to track discussion on that.

### DIFF
--- a/software/python/ayab/ayab.py
+++ b/software/python/ayab/ayab.py
@@ -233,7 +233,7 @@ class GuiMain(QMainWindow):
         '''Updates the current scene '''
         width, height = self.pil_image.size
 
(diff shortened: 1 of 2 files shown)


chris007de:
Thanks, I'll have a look at #173

PR #178, opened 2017-02-22
Remove extraneous 'e' on except case

oskay:
Can cause a crash with "NameError: global name 'e' is not defined"


chris007de:
@oskay I forgot that this is already fixed in the v0.9 branch: https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/v0.9/software/python/ayab/plugins/ayab_plugin/firmware_flash.py#L166
The exception is written to the log to ease debugging.

PR #182, opened 2017-03-01
Fix two windows that failed to open, bump version number

oskay:
Both the "smart resize" and the "about" window were failing to open. In
both cases, QtGui-> QtWidgets fixes the issue.

Also bump the version number display to 0.9, since this is the 0.9
branch.


chris007de:
Thx! 🎉

PR #265, opened 2018-06-07
Fix rendering on High DPI displays (e.g. MS Surface and 4K displays in Windows)

caelor:
This has been tested in a development environment on Windows, but the code is based on example code from https://github.com/x64dbg/PyQt5/wiki/Troubleshooting so I'm confident it shouldn't introduce regressions on other platforms.

I've put the README.md update in a separate commit so it can be separated if a different update is preferred. The official Windows builds will need to use PyQt 5.6 or above in order for the code to work (according to http://doc.qt.io/qt-5/highdpi.html)


chris007de:
Thanks a lot! Will test and merge asap.

Issue #270, opened 2018-11-24
Limit multiplier in repeat rapport menu and handle excessive pattern widths


chris007de:
> We could limit the multiplier to ceil(200/patternWidth). So it is possible to multiply until you are exceeding 200 needles, but can't enter completely random values.

Issue #238, opened 2017-09-11
Save a set of user preferences


Adrienne200:
When working on a project, I'll be using the same settings every time I open an image. It would be nice if it remembered a set of Favorites or Preferences so I don't have to set them every time.
 - mirrored or not
 - 3 or more colors
 - single/ribber/circular
 - infinite repeat
 - alignment
 - anything else?
And a Reset to put them all back to default.

chris007de:
As we are operating on multiple platforms, we need a versatile solution for this.
Portability/backupability would be a plus, so just storing a .ini file in the user folder would probably a  viable solution (configparser?).
Alternatively, PyQt provides QSettings, which handles storage of settings transparently (registry for Windows, hidden config file for Linux).
Aside from how it is solved under the hood, I propose to implement a helper class as an abstraction layer (AYABPreferences).

t0mpr1c3:
I can have a go -- assign to me?

chris007de:
Tested on Windows 10 with https://github.com/AllYarnsAreBeautiful/ayab-desktop/actions/runs/140183655:
Infinite Repeat and Mirroring are shown as enabled on first start of the software. When altering the values in the Preferences dialogue, the values are written to the Registry. After restarting the software, they are checked again, so it seems that they are not properly read from Registry.
The values for Alignment and Knitting mode are working as expected, so only the boolean values are having this issue.

t0mpr1c3:
I remember having a problem with boolean values being written to `ayab.conf` in lower case ("true","false") but on reading are expected to be title case ("True","False"). Can you check what values are actually written to the registry?

chris007de:
Every value in the registry is written lowercase.

chris007de:
Still have to test if this is working properly on Windows and macOS now.

t0mpr1c3:
Works OK on Windows, I think. I just made a small change to catch type errors in the saved settings (https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/378).

t0mpr1c3:
Chris, I can run some checks on a Mac in a couple of weeks when I will have access to a MacBook.

t0mpr1c3:
I suggest that this issue can be closed if Mac users are not reporting any issues.

t0mpr1c3:
Are there outstanding issues, or can this be closed?

PR #309, opened 2020-06-04
implemented Preferences class and menu item

t0mpr1c3:
With reference to https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/238. Implemented using QSettings.

Commit 50be007 deals with https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/270.


chris007de:
Hi Tom, it looks like you pushed two additional commits to the branch where you developed this Pull Request ("implemented ceiling for horizontal repeat ..."). Github watches the branch where a PR originates, automatically detects further additions and adds additional commits to the PR.
This is useful when a PR is still work in progress, but in your case, it polluted the PR with another topic.
We'll ignore this for now, but for the future it would be good if you maintain separate branches for your Pull Requests to avoid such situations, thanks a lot :)

Issue #308, opened 2020-06-02
Internal colour representation


t0mpr1c3:
As Chris mentioned (https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/297) the internal concept of colour, as implemented in ayab.plugins.ayab_plugin.ayab_image, is represented as bands of greyscale brightness.

I haven't tested it, but presumably this would not distinguish between e.g. cyan 0xFFFF00, magenta 0xFF00FF, and yellow 0x00FFFF,

It probably works fine most of the time but it does look like the type of kludge that we should probably get away from before releasing a 1.0.0 version.

Chris's suggestion of a hard limit of 4 or 6 colours seems sensible, and these few colours could be represented by integer values just as they are now, with a simple lookup table for the RGB values.

More generally, I am concerned that the reliance on a one-byte-per-stitch matrix (plus a few knitting modes) as a representation of the knitting pattern essentially restricts AYAB to rectangles of stocking stitch, in contrast to more flexible formats discussed elsewhere e.g. https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/299. Changing the internal representation wholesale would clearly be a massive refactoring, but as I mentioned in that thread, a more abstract representation would offer considerable advantages.

chris007de:
I am open for a versatile concept for this. The AYABImage class has to be refactored anyway, so we can use this as a chance to make it future-proof as well.

t0mpr1c3:
What's the most useful way I can contribute at the moment? I've been adding little features, just because that is fairly easy to do, but I would be happy to attempt some of the heavier tasks.

chris007de:
> What's the most useful way I can contribute at the moment? I've been adding little features, just because that is fairly easy to do, but I would be happy to attempt some of the heavier tasks.

Probably we should handle #299 first, so we have a consolidated set of requirements for a holistic pattern representation, i.e. which properties of a pattern have to be handled inside the software (colors, shape, carriage, instructions, ...). Then we can dive into re-implementation of AYABImage (perhaps AYABPattern by then).
As you are definitely closer to the topic of knitting and you already made up your mind a bit about a knitting pattern format, I propose that you document some of your thoughts and (high level) requirements into #299. What do you think?

t0mpr1c3:
Yes, I think that's a good suggestion.

t0mpr1c3:
[...]

PR #313, opened 2020-06-10
Altered image to use palette of colors instead of greyscale

t0mpr1c3:
Implementing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/308


chris007de:
Nice and clean solution :)

Issue #301, opened 2020-04-10
Refactor/Implement unit tests for ayab_control::__cnfLine


chris007de:
To start out with the refactoring of the code and implementation of unit tests, we should make the cnfLine() method testable, as it is the heart of the software and yields a lot of possibilities for bugs and malfunctions.

t0mpr1c3:
__cnfline() really needs to be broken up into smaller component functions. These would be much easier to test, and would facilitate code reuse.

PR #322, opened 2020-06-12
refactored __cnfLine

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/301#issuecomment-642934667

Knitting mode functions moved out of __cnfLine into separate methods. Algebra simplified here and there. Simulated output tested identical for every knitting mode. Could use a run through yapf.


chris007de:
Looks good.
Would be great to have tests for the knitting functions added, though - or do you want to handle testing in another PR?

t0mpr1c3:
Yes, I was thinking I'd do that next.

Issue #311, opened 2020-06-09
Implement CRC8 for cnfLine message


dl1com:
add a CRC8 on bytes 0-28 of the cnfLine message, so the Arduino can check if the line data is correct and re-request in case of failure (https://github.com/AllYarnsAreBeautiful/ayab-firmware/blob/v1.0-dev/src/ayab/serial_encoding.cpp#L129)

t0mpr1c3:
Currently the device does not signal checksum errors, or any other kind of error, in cnfLine.

Nor does the host check for an error and repeat the reqLine message if an error is detected.

dl1com:
Implementation is done, closing.

PR #323, opened 2020-06-12
CRC8 check

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-firmware/issues/6 and https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/311.

Hopefully this should allow testing of the CRC8 check in the firmware.

### DIFF
--- a/src/main/python/ayab/tests/test_communication.py
+++ b/src/main/python/ayab/tests/test_communication.py
@@ -86,11 +86,11 @@ class TestCommunication(unittest.TestCase):
         assert bytes_read == byte_array
 
     def test_cnf_line(self):
-        lineNumber = 13
-        lineData = [0x23, 0x24]
-        flags = 0x12
-        crc8 = 0x57
-        self.comm_dummy.cnf_line(lineNumber, lineData, flags, crc8)
+        lineNumber = 0
+        lineData = b'\0\xde\xad\xbe\xef\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0'
+        flags = 1
+        crc8 = 0xE9
+        self.comm_dummy.cnf_line(lineNumber, lineData, flags)
         byte_array = bytearray([0xc0, 0x42])
         byte_array.append(lineNumber)
         byte_array.extend(lineData)
--- a/src/main/python/ayab/plugins/ayab_plugin/ayab_communication.py
+++ b/src/main/python/ayab/plugins/ayab_plugin/ayab_communication.py
@@ -135,7 +135,7 @@ class AyabCommunication(object):
         data = self.__driver.send(b'\x04')
         self.__ser.write(data)
 
-    def cnf_line(self, lineNumber, lineData, flags, crc8):
+    def cnf_line(self, lineNumber, lineData, flags):
         """Sends a line of data via the serial port.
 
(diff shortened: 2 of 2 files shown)


src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
```diff
@@ -415,14 +416,18 @@ def __cnfLine(self, lineNumber):
                     if 0 <= pxlNumber and pxlNumber < 200:
                         self._set_pixel(bytes, pxlNumber)
 
-            # TODO implement CRC8
-            crc8 = 0x00
+            # calculate CRC8 hash
+            hash = crc8.crc8()
+            hash.update(bytes)
+            hash.update(bytearray([reqestedLine]))
```
chris007de, line 422: Perhaps you could move this into a separate method which takes the bytes from the whole cnfLine packet and call it in Line 426 and 429. This would enable testing of the method to ensure it yields the same output as in the firmware (e.g. with the same test data as in https://github.com/AllYarnsAreBeautiful/ayab-firmware/blob/v1.0-dev/test/test_serial_encoding.cpp#L49-L55)

t0mpr1c3:
Sounds good, Chris. I was actually thinking of removing the CRC byte from the arguments for cnf_line() making the CRC code internal to ayab_communication.py. Does that make sense?

chris007de:
> I was actually thinking of removing the CRC byte from the arguments for cnf_line() making the CRC code internal to ayab_communication.py. Does that make sense?

Yes, definitely a good idea :)

src/main/python/ayab/tests/test_communication.py
```diff
@@ -90,7 +90,7 @@ def test_cnf_line(self):
         lineData = [0x23, 0x24]
         flags = 0x12
         crc8 = 0x57
-        self.comm_dummy.cnf_line(lineNumber, lineData, flags, crc8)
+        self.comm_dummy.cnf_line(lineNumber, lineData, flags)
         byte_array = bytearray([0xc0, 0x42])
```
t0mpr1c3, line 94: I calculate the checksum as 0x88.

chris007de: We have some discrepancy here.
@t0mpr1c3  has used a vanilla crc8 implementation, whereas @Sturla22 has used the Maxim algorithm for ayab-firmware. These algorithms yield different results for the same input vector. You can compare the outputs at https://crccalc.com/

The vector used by Sturla in the ayab-firmware test (https://github.com/AllYarnsAreBeautiful/ayab-firmware/blob/v1.0-dev/test/test_serial_encoding.cpp#L51) yields 0xE9 with CRC8/maxim, and yields 0x3D with vanilla CRC8.
So we have to change one of the algorithms ;) Who wants to? :)

```
0x42 0x00 0x00 0xde 0xad 0xbe 0xef 0x00  0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00
0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00  0x00 0x00 0x00 0x01
```

t0mpr1c3: I should match Sturla's, I suppose! Apart from anything else, there are fewer constraints on the desktop software.

Issue #257, opened 2018-03-14
Continuous status reporting


dl1com:
To be able to show more information about the current machine status and for improved possibilities for debugging, the machine shall transmit some information about its current state.
Currently, this is only done in "Test Mode" (no knitting possible), so the continuous data transmission does not interfere with the knitting process.
But I guess it should be feasible to knit and send data in parallel.
To do so, the message handling on both sides (AYAB GUI and Arduino Firmware) has to be reviewed and optimized.

The information we want send to the computer is:
- Type of carriage (K, L, G)
- Direction of movement
- "Initialized?"-State
- Carriage position

Data we could try to send also:
- Value of hall sensor measurement, e.g. to be able to debug broken magnets or hall sensors.

oskay:
I'd suggest that the carriage position indicate both row and column.

chris007de:
Good point. Perhaps I would extend it to count also the rows since the last hall sensor pass, to take debugging of #161 into account.

Adrienne200:
So far so good, this will be a very nice feature, but it is still very slow (tested with post61, for 0.95). Move the carriage across and then you have to wait about 8 seconds while the carriage display catches up with where it actually is. Then it finally beeps and clicks up to the next row. Could it check every second or two where the carriage is now and just jump there?
The carriage display isn't centered. Stop the carriage with the magnet centered, on or between L1 and R1, and wait for the display to catch up. It should be in the middle of the graphic, and it isn't.
Would be nice to have numbers on the graphic, or show the position number somehow. Maybe just when the carriage comes to rest.
A direction indicator would be nice, an arrow down below the row number.

chris007de:
I would propose to leave it as it is for now and do improvements in the next version as this is not a core feature and we are running out of time this year.
The remaining question is whether to enable people to use it in the current state in 0.95 or to hide the checkbox for this mode for now.
What do you think?

Adrienne200:
The Status tab is still visible but is always grey. Did you want to suppress it too?

glindsey:
Just wanted to note: I'm trying to debug our current AYAB KH-910 installation, and this feature would be _really_ helpful for that, so I'd love to see it in the software eventually. :)

glindsey:
[...]

Issue #226, opened 2017-07-21
Include the sub-row numbers and color-changer button number when knitting DBJ


Adrienne200:
When knitting DBJ and circular it take multiple passes to knit one row. Currently it just displays the same row number until that row is finished.
It would be nice if it displayed which pass you're on, and which color-changer button.
For example, 3-color DBJ which takes 6 passes per row:
 Row - Pass - CC
1 - 1 - 1
1 - 2 - 1
1 - 3 - 2
1 - 4 - 2
1 - 5 - 3
1 - 6 - 3
That's rather cryptic, maybe there's a better design, but that's the idea.

Adrienne200:
This is mostly implemented now, (0.95 RC3) thanks. But some ideas:
 - when using single bed, it says "Color A/B" all the time. While this is true, I'm not sure it adds much value. Most people do the majority of their knitting single bed, and having this text sitting there all the time might be considered clutter. Don't know, maybe it's OK.
 - it's very very useful with DBJ though. At first I found the time where it changes color to be a bit of a surprise, but actually it's correct, and essentially the same as Brother did it. For 2-color classic, as you turn around on the right it switches to telling you which color comes next, and stays there until you're on the right again. So on the way into the changer it's telling you which button to press, and as you leave you can confirm that you did it right. Perfect, exactly what I wanted.
But on the various 3-color DBJ's, it often says "Color: Blank" This has some value for understanding what the selection is doing (and probably for debugging) but I think this field should be only for the actual color that will be changed next, (or has just been changed to) regardless of whether it's doing a selection or non-selection row. While testing I often had to wait for the next beep to confirm which color I needed next.
 - at the very beginning of a piece, it sometimes displays the wrong color. I think it doesn't know for sure which color until after the triple-beep when you start. Better to display nothing than a random color left over from the last pattern. Maybe it could clear this field when you click Configure? Or if it does know, can it display it sooner?

chris007de:
> - when using single bed, it says "Color A/B" all the time. While this is true, I'm not sure it adds much value. Most people do the majority of their knitting single bed, and having this text sitting there all the time might be considered clutter. Don't know, maybe it's OK.

Let's remove it.

[...]

PR #334, opened 2020-06-17
Knit progress dialog

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/226 and https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/257

A window opens displaying row, color, and needle selection information as knitting proceeds.

It will need a good amount of testing to refine the information displayed and check that the UI is robust. I think there is something wrong with how it re-initializes because the dialog box works pretty well first time around, at least in the simulation mode, but there are frequent glitches when restarting the knitting process.


chris007de:
Thanks a lot for another useful addition to the project!
I love the representation of the colors, that's really a huge improvement.

Some thoughts on it:
- Why a separate Window?
  - When clicking "Configure", it steals the attention from the main window, but we still have to click "Knit" to actually start knitting. That's not very intuitive.
  - The window shows also when I configure a non-allowed setting (singlebed with 3 colors)
  - Wouldn't it also be OK to split the main window horizontally and show the content of the progress window there? (e.g. Top: Overview, Bottom: Knitting Progress)

- Automatic scrolling is a must-have, I guess (at least on my system, it does not scroll automatically on addition of a new pass)
- When I cancelled knitting and start another knitting, the program is stuck . It seems that the knitting state (reqLine etc.) are not reset properly on cancel. (When cancelling at row 2, the first requested row at the next knitting is row 3). Not sure if related to the progress window, though.
- Also not sure if this is related to the progress window: In multicolor, if we do not use a certain color in a row, couldn't we skip the passes for this color?
- What's the meaning of the small triangles inside the color blocks?

src/main/python/ayab/ayab.py
```diff
@@ -48,6 +52,12 @@
 
 MACHINE_WIDTH = 200
 
+# TODO move to generic configuration
+MACHINE_WIDTH = 200
```
chris007de, line 56: Duplicate of line 53

src/main/python/ayab/ayab.py
```diff
@@ -65,18 +75,51 @@
...
+
+def rgb2array(color):
+    return np.array([color // 0x1000, (color // 0x100) & 0xFF, color & 0xFF])
+
+def greyscale(rgb):
+    return np.dot([.299, .587, .114], rgb)
+
+def contrast_color(color):
+    return [0x000000,0xFFFFFF][greyscale(rgb2array(color)) < 0x80]
+
```
chris007de, line 110: Probably worth to put these functions as methods into the ayab_progress class.

src/main/python/ayab/plugins/ayab_plugin/ayab_communication_mockup.py
```diff
@@ -54,6 +58,20 @@ def update(self) -> list:
             self.__line_count += 1
             self.__line_count %= 256
             self.__rxMsgList.append(reqLine)
+            if self.__delay:
+                sleep(1) # wait for knitting progress dialog to update
+
+            # nasty kludge: set condition to True for line-by-line debugging
+            if False:
```
chris007de, line 65: Maybe decide this with a parameter in the constructor, so we can choose the behavior with a parameter in the GUI in the future.

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
```diff
@@ -472,46 +537,46 @@ def knit(self, pImage, pOptions):
...
-        elif self._current_state == KnittingState.INIT:
+        elif self.__current_state == KnittingState.INIT:
             if rcvMsg == 'cnfInfo':
-                if rcvParam == self._API_VERSION:
-                    self._current_state = KnittingState.WAIT_FOR_INIT
+                if rcvParam == self.__API_VERSION:
+                    self.__current_state = KnittingState.WAIT_FOR_INIT
                     result = AYABControlKnitResult.WAIT_FOR_INIT
                 else:
                     self.__logger.error("wrong API version: " +
```
chris007de, line 548: Unrelated: Sooner or later, we should show the Firmware Version here, as the user is not aware of an API version at all.

src/main/python/ayab/ayab.py
```diff
@@ -160,6 +213,41 @@ def updateStatus(self, hall_l, hall_r, carriage_type, carriage_position):
...
+                # TODO: hints, notes, memos
+                # self.kp.area.ensureWidgetVisible(w0)
+            # graph line of stitches
+            for c in range(len(progress.bits)):
+                wc = stitch(progress.color, 0, progress.bits[c], progress.alt_color)
+                self.kp.grid.addWidget(wc, self.kp.row, 6 + c)
+            # if not infinite repeats, update progress bar
+            if progress.repeats < 0 and progress.total_rows > 0:  
+                self.kp.pd.setValue(100 * (row + 1) / progress.total_rows)
+            self.kp.pd.update()
```
chris007de, line 249: Move logic into ayab_progress method?

t0mpr1c3: Yes

src/main/python/ayab/ayab.py
t0mpr1c3: Thanks. Eventually I'm going to move all the constants into a single module as you suggested.

src/main/python/ayab/ayab.py
t0mpr1c3: Yes

src/main/python/ayab/plugins/ayab_plugin/ayab_communication_mockup.py
t0mpr1c3: Yes

t0mpr1c3:
Thanks Chris -- these are very helpful comments

1. It probably should not be in a separate window, but I don't have a very good sense of how to organize the viewing area at the moment. The horizontal split is a sensible suggestion. Probably it should go at the bottom near where the progress bar is at the moment. And it should probably emerge when you press "Knit" instead of "configure".

2. I struggled with implementing automatic scrolling. I will figure it out eventually.

3. Yes, this is a particularly bad bug. Something does not reset properly on cancel.

4. Yes, as far as I can tell this is missing from the knitting logic entirely. The only suggestion that I have seen is not to design patterns with bank rows! (I think that's in one of the DBJ manual pages.) I suggest that we raise a new issue to deal with that.

5. It's an experimental feature, we can get rid of them. The little dots are intended to represent knit stitches (as opposed to purl stitches which are usually charted with a vertical line, yarn overs with a circle, etc). Since we do not currently implement any knowledge of different stitch types there is no reason to show these symbols at the moment. (I would like to change that in future so that it is possible to knit in both different colours and different textures. At the moment AYAB barely supports lace knitting. But it will be a big change, and I am sure it won't happen any time soon.)

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
```diff
@@ -181,88 +215,94 @@ def __cnfLine(self, lineNumber):
...
 
-            # TODO implement CRC8
-            crc8 = 0x00
-
             # send line to machine
-            if self._infRepeat:
-                self.__ayabCom.cnf_line(requestedLine, bits.tobytes(), 0, crc8)
-            else:
-                self.__ayabCom.cnf_line(requestedLine, bits.tobytes(), lastLine, crc8)
+            self.__ayabCom.cnf_line(requestedLine, bits.tobytes(), lastLine & self.__infRepeat)
```
chris007de, line 225: AyabCommunicationMockup.cnf_line() still expects (lineNumber, lineData, flags, crc8). Please remove the crc8 parameter so we have the same interface for both variants

t0mpr1c3: OK

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
```diff
@@ -54,121 +60,136 @@ class AYABControlKnitResult(Enum):
...
-        elif self._knitting_mode == KnittingMode.MIDDLECOLORSTWICE_RIBBER.value and self._numColors >= 2:
-            self._knit_func = self._middlecoltwice
-        elif self._knitting_mode == KnittingMode.HEARTOFPLUTO_RIBBER.value and self._numColors >= 2:
-            self._knit_func = self._heartofpluto
-        elif self._knitting_mode == KnittingMode.CIRCULAR_RIBBER.value: # not restricted to 2 colors
-            self._knit_func = self._circular_ribber
-        else:
-            self.__logger.error("Fallthrough error in __get_knit_func: invalid knitting options")
-            return False # knit function not found
+        if not self.__knitting_mode.good_ncolors(self.__numColors):
```
chris007de, line 164: self.__knitting_mode needs a cast to KnittingMode().

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
```diff
@@ -181,88 +202,83 @@ def __cnfLine(self, lineNumber):
...
             lineNumber += BLOCK_LENGTH * self.__lineBlock
 
-            imgHeight = self._image.imgHeight()
-        
-            # work out which line of data to send
-            imgRow, color, indexToSend, sendBlankLine, lastLine = self._knit_func(lineNumber, imgHeight)
-            
-            # create bitarray
+            # get data for next line of knitting
+            imgRow, color, indexToSend, sendBlankLine, lastLine = self.__knit_func(lineNumber)
```
chris007de, line 206: Trying to knit singlebed yields: `TypeError: _singlebed() missing 1 required positional argument: 'lineNumber'`.
Possible to add a unit test for cnf_line?

t0mpr1c3:
Thanks, Chris! I will get around to the making unit tests for the new methods at some point. Right now I am trying to fix up the UI. I switched from a pop-up dialog box to a split window as you suggested. It needs a few tweaks but overall I think it is much better. Also I think I have fixed the problem I was having with the auto scrolling. It still crashes when you try to knit again after cancelling, I am trying to track that down next.

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
t0mpr1c3: Done

src/main/python/ayab/plugins/ayab_plugin/ayab_control.py
t0mpr1c3: Don't we already have tests for `cnf_line()` in `test_communication.py` and `test_communication_mockup.py`?

t0mpr1c3:
Chris -- is restarting knitting where you left off a bug, or is it supposed to be a feature? If it is a feature, I suggest that it should be implemented by setting the "start line" form entry to the last line that was actually completed plus one.

chris007de:
> Chris -- after cancelling knitting, is restarting where you left off a bug, or is it supposed to be a feature?

As far as I remember, it should start from scratch and not preserve the line - so I would call it a bug (because we do not have a UI element to actually reset it to start from scratch, so it should happen automatically)

t0mpr1c3:
I started a new issue to address the program flow bugs: https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/338

t0mpr1c3:
I think I fixed the bug. The problem was that after configure, knit, cancel, configure, knit, the FSM was never reset to SETUP and was still in OPERATE.

t0mpr1c3:
Chris, this is ready for further review when you get a chance.

chris007de:
Great work...only one drawback: The unit tests (especially test_req_line) currently take about 4:30 minutes to complete. We should definitely address that :)
I'll merge this for now, polishing/issues can be addressed separately.
Thanks a lot!

t0mpr1c3:
Thanks. I have no idea why the tests are so slow.

Issue #1, opened 2016-05-25
put DROPS knittig instructions into wiki


niccokunzmann:
[Here](http://www.garnstudio.com/pattern.php?id=7410&cid=9) we can find sets of knitting instructions. 
These should be put into the Wiki or a table where we can list all of them. They are the basis of the knit files.

![](http://www.garnstudio.com/drops/mag/170/24/24-diag.jpg)

niccokunzmann:
You can view the knitting instructions [here](https://github.com/AllYarnsAreBeautiful/ayab-desktop/wiki/2016-05-25---Knitting-pattern) Be aware taht these are not all instructions as knitting can never be fully covered by instructions.

Issue #345, opened 2020-06-20
Remove the Configure button


t0mpr1c3:
Could we get rid of the configure button, and just automatically configure when the knit button is pressed?

chris007de:
> Could we get rid of the configure button, and just automatically configure when the knit button is pressed?

Definitely :) I don't see any benefits from it.

t0mpr1c3:
Should be quite a straightforward change. First task was to straighten out the FSM which I did in https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/347.

Issue #338, opened 2020-06-18
Two finite state machines?!


t0mpr1c3:
There is one FSM specified in https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/1.0.0-dev/src/main/python/ayab/plugins/knitting_plugin.py and another one defined by `class KnittingState(Enum)` in https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/1.0.0-dev/src/main/python/ayab/plugins/ayab_plugin/ayab_control.py

The second one seems to have been introduced more recently. Can we get rid of it? It's confusing and complicates the program flow.

I suspect that doing so will help get rid of the various crashes and glitches that tend to occur after hitting Cancel, for example https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/334#issuecomment-645445528.

chris007de:
They do not serve the same purpose, so they are not redundant.
The state machine in knitting_plugin is a high level FSM which accounts the overall state of the software (Configured, Knitting). The second FSM in ayab_control has recently been refactored, as it has been always there but was not really recognizable as an FSM. It controls the communication between the host and the Arduino while in the (super)state Knitting.
I agree that this can be optimized - probably we can also get rid of the FYSOM dependency, as building an FSM probably does not need a library ;-)

t0mpr1c3:
If you count the Qt event handling machinery, there are arguably 3 FSMs going on ;-) I wonder if that is actually where the problem lies.

I'm not necessarily against `fysom`. It has the advantage that the behaviour is specified in a declarative way, instead of having to follow a tangle of `if .. else` clauses.

I'm also not against having a FSM for the sole purpose of handling the serial communication if it is well encapsulated. If you have recently refactored it, then it is probably in good shape -- I admit that I haven't followed all of the logic.

t0mpr1c3:
The problem was that the FSM in `ayab_control.py` was not being reset after cancelling, and stayed in OPERATE. I've prevented that by making a new method `ayab_control.reset()` that sets the state back to SETUP. That seems to have fixed it for now, but I haven't tested exhaustively.

t0mpr1c3:
Do you think the Qt Finite State Machine framework might be useful instead of `fysom`? We have state changes triggered by button presses. And it would be one less dependency.

chris007de:
[...]

PR #347, opened 2020-06-22
migrated from Fsyom to QStateMachine

t0mpr1c3:
* migrated from Fsyom to QStateMachine
* deleted empty plugins
* removed race condition by adding FINISHING State
* fixed menu behaviour
* moved graphics scene methods to separate module


t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/338

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/345#issuecomment-646997698 (Configure button)

t0mpr1c3:
simpleaudio requires asoundlib.h (`apt install libasound2-dev`)

t0mpr1c3:
Chris, this is ready to go now that I have fixed the audio library dependency.

src/main/python/ayab/ayab_fsm.py
```diff
@@ -0,0 +1,127 @@
...
+class FSM (object):
+    """Heirarchical Finite State Machine.
+
+    This replaces `fsyom` and various imperatively-programmed structures
+    with a declaratively-programmed, event-driven framework tailored to the UI.
+    @author Tom Price
+    @date   June 2020
+    """
+    def __init__(self, parent):
+        """Define Heirarchical states for Finite State Machine"""
```
chris007de, line 36: typo

src/main/python/ayab/ayab.py
```diff
@@ -549,14 +358,19 @@ def getSerialPorts(self):
...
-        # if event == "start":
-        #     playsound(self.app_context.get_resource("assets/start.wav"))
-        # if event == "nextline":
-        #     playsound(self.app_context.get_resource("assets/nextline.wav"))
-        # if event == "finished":
-        #     playsound(self.app_context.get_resource("assets/finish.wav"))
+    def playsound(self, sound):
+        if str2bool(self.prefs.settings.value("quiet_mode")):
+            return 
+        dirname = path.join("..", "..", "resources", "base", "assets", path.dirname(__file__))
```
chris007de, line 364: When running the application, the .wav files are not found. (`WARNING  File start.wav not found.`)
Please use self.app_context.get_resource() (see before your commit) to get an OS independent location of the files.

src/main/python/ayab/ayab.py
```diff
@@ -549,14 +358,19 @@ def getSerialPorts(self):
...
-        # if event == "start":
-        #     playsound(self.app_context.get_resource("assets/start.wav"))
-        # if event == "nextline":
-        #     playsound(self.app_context.get_resource("assets/nextline.wav"))
-        # if event == "finished":
-        #     playsound(self.app_context.get_resource("assets/finish.wav"))
+    def playsound(self, sound):
+        if str2bool(self.prefs.settings.value("quiet_mode")):
+            return 
+        dirname = self.app_context.get_resource("base/assets")
```
chris007de, line 364: Needs to be: `dirname = self.app_context.get_resource("assets")`
Could have told you that in my first comment already

chris007de:
* For me, the progress window always starts in the 2nd or 3rd row. Console log shows correct rows.
* Additionally, cancelling knitting and switching to another pattern shows the previous pattern in the progress window.

t0mpr1c3:
> * For me, the progress window always starts in the 2nd or 3rd row. Console log shows correct rows.
> * Additionally, cancelling knitting and switching to another pattern shows the previous pattern in the progress window.

1. I set a 2 second delay in simulation mode before knitting starts to try to avoid that, but it would be better to require an acknowledgement signal.
2. The knit progress continues to show after cancel has been pressed, but does go away when Knit is pressed a second time. Do you think it should disappear sooner than that?

chris007de:
> > * For me, the progress window always starts in the 2nd or 3rd row. Console log shows correct rows.
> > * Additionally, cancelling knitting and switching to another pattern shows the previous pattern in the progress window.
> 
> 1. I set a 2 second delay in simulation mode before knitting starts to try to avoid that, but it would be better to require an acknowledgement signal.

Ok, so I understand that it is an issue of the simulation mode only. We'll address this in a separate issue.

> 2. The knit progress continues to show after cancel has been pressed, but does go away when Knit is pressed a second time. Do you think it should disappear sooner than that?

The way the progress data appears and disappears is fine for me. What I wanted to say is that if I knit pattern A, cancel, load pattern B, knit, still the data of pattern A is shown in the progress window. Please try if you can reproduce this on your side.
we can address this in a separate issue as well.

t0mpr1c3:
Thanks, I implemented the `get_resource()` call. I haven't been able to reproduce the second problem yet.

t0mpr1c3:
I fixed bug #1 by synchronizing the threads at the "Please Knit" notification. I still can't reproduce bug #2 at the moment.

Issue #281, opened 2019-01-09
Maintain translation files for German


chris007de:
The translation files for German are badly out of date. Some labels in the UI are not prepared for translation.

t0mpr1c3:
I'm afraid that my recent additions have not helped matters, because the new UI stuff is written as Python instead of Qt XML.

chris007de:
I'd say this is solved by the new translator feature by @t0mpr1c3

PR #352, opened 2020-06-25
Updated translation files

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/281

The translations have mostly copied over from the previous versions. There are a couple of additions courtesy of Google Translate that need checking from a native speaker!

I have also had a go at the notifications in `ayab_plugin.py`. These definitely need to be reviewed.

I edited the `.gitignore` file to disregard `.qm` files, which are binaries. The build process should be changed to generate these using `lrelease`, if it does not already do that.


t0mpr1c3:
Referencing issue #353

chris007de:
> I edited the `.gitignore` file to disregard `.qm` files, which are binaries. The build process should be changed to generate these using `lrelease`, if it does not already do that.

The .qm files are not part of the build environment, yet. I fully support to remove the binaries from version control and to generate them during the build.
But: For (linux) users which want to use the software right from a cloned git repository, we should at least add the necessity for manual generation of these files before being able to run with `fbs run` to the documentation.

t0mpr1c3:
Done: add language preference, needs debug. To do: add Italian, update German.

t0mpr1c3:
The script that generates the translation files is trivial, but it is written in Perl which creates a new dependency. I suppose I should have done it in Python? I didn't really think about it because just about every OS has Perl bundled in.

.github/workflows/build.yml
```diff
@@ -37,6 +37,10 @@ jobs:
         pip install -r windows-build\windows_build_requirements.txt
     - name: Build app
       run: |
+        cd src/main/resources/base/ayab/translations
+        ./ayab_trans.pl
+        lrelease *.ts
+        cd ../../../../../..
```
chris007de, line 43: Should be possible to simply cd back to $HOME (or ~) (see https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)

chris007de:
Running the application fails with 

```
ayab.py", line 428, in run
    translator.load("ayab_trans." + language, lang_dir)
TypeError: can only concatenate str (not "NoneType") to str
```

So there was no language found, but also, no exception was raised by `QSettings().value("language")`

.github/workflows/build.yml
t0mpr1c3: OK. I made a new workflow step to create the translation files. I'm not familiar with these workflows, so I hope that works.

t0mpr1c3:
OK, I made some changes to the error handling. If the preferred language is not available it should use the default locale or "en_US".

t0mpr1c3:
The last thing that needed doing was translation for combo box items. The simplest way was to create separate classes for each combo box.

t0mpr1c3:
The final commit includes a change to `test_communication_mockup.py` so that the test of `reqLine()` takes a few seconds instead of a few minutes.

PR #373, opened 2020-07-08
More changes towards a modular class structure

t0mpr1c3:
Renamed AyabImage to AyabPattern, in preparation for new AyabImage class.

    Created AyabImage class as container for Scene.image and image action methods.

    Defined KnitProgress as subclass of QScrollArea.

    Renamed signal classes to AyabObserver and AyabObservable, after observer
    pattern, with Observable class used by inheritance rather than composition.

    Defined Scene as subclass of QGraphicsView.

    Subsumed Progress class into ProgressBar.

    Renamed states in finite state machine.


t0mpr1c3:
Thanks, Chris. I rewrote the preferences class and dialog so that it should be easily extensible going forward: all you need to do is add the name and Class to a dict variable.

Issue #286, opened 2019-01-10
The icon doesn't show well in the Windows10 task bar


Adrienne200:
Pin the AYAB icon to the task bar. While the app is running, the icon is in a gray box and shows up well enough. But when it's not running it's black on black, barely visible.
<img width="539" alt="icon in taskbar win10 ayab open" src="https://user-images.githubusercontent.com/24358225/50952438-4ac22280-1465-11e9-8b77-0971c3f09265.PNG">
<img width="377" alt="icon in taskbar win10 ayab closed" src="https://user-images.githubusercontent.com/24358225/50952650-ec497400-1465-11e9-8a8e-464f6852bca6.PNG">

chris007de:
Perhaps somebody in the community could take care of this...some rework on the icon (visible on dark backgrounds and higher resolution?)

derchrisuk:
Invert colors ?

chris007de:
@derchrisuk good idea...or applying some grey shadows?

PR #376, opened 2020-07-10
nicer icon maybe

t0mpr1c3:
Same ball-of-yarn concept, but I tried to make it look like a letter "A", for AYAB.

Also, it shows up OK on white and black backgrounds (ref. https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/286).


t0mpr1c3:
icon

Issue #208, opened 2017-06-30
Play sound files in case of start, end of line, end of knitting


chris007de:
Instead of using the beeper, play sound files on the computer to indicate the start of knitting, end of line and end of knitting to the user.

chris007de:
https://pypi.org/project/playsound/

chris007de:
playsound library causes problems on Windows and macOS, try https://simpleaudio.readthedocs.io/en/latest/

chris007de:
Currently (https://github.com/AllYarnsAreBeautiful/ayab-desktop/commit/6ede3488f6c03b10574e950c1f8c05774707c785), we are only playing a sound when starting and when finished.
I know from some users who do not like the hardware beeper (or it is too quiet), so being able to have an (optional) sound on each line would be desirable, I would say.

t0mpr1c3:
Looks like we have a new line sound ready to go :) It is a bell sound like an old-fashioned typewriter.

I can implement this easily enough. We already have a "quiet mode" setting that turns off all the sounds, do we need an extra setting just for the new line sound?

chris007de:
> I can implement this easily enough. We already have a "quiet mode" setting that turns off all the sounds, do we need an extra setting just for the new line sound?

I think it is not necessary to control this separately. So when not quiet, I'd play all sounds (start, newline, finish). Let's see how many people get annoyed by this ;)

t0mpr1c3:
That sounds fine to me.

I might create a new thread to play the sounds to avoid blocking the UI thread while sounds are playing.

PR #382, opened 2020-07-12
Added audio player thread and new line sound.

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/208#issuecomment-657250652

Added bicycle-like bell sound after sending each line, except after the final line when transmission has finished. I am expecting that when we release the new version, 50% of all the feedback we get will concern this single feature. ;)

Created a new audio thread for the sound player. I'm not sure this is absolutely necessary because the connection to it is blocking anyway, but it provides additional options for resource sharing.

### DIFF
--- a/src/main/python/ayab/ayab.py
+++ b/src/main/python/ayab/ayab.py
@@ -21,19 +21,16 @@
 from fbs_runtime.application_context.PyQt5 import ApplicationContext
 
 import sys
-from os import path
 import logging
 
 from PyQt5.QtWidgets import QMainWindow, QApplication, QMessageBox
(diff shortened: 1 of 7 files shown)


chris007de:
Hmmm I probably merged this too early...when running the application, exits with:

```
File "/ayab-desktop/src/main/python/ayab/ayab_observer.py", line 69, in activate_signals
    type=Qt.BlockingQueuedConnection)
TypeError: argument 1 has unexpected type 'AudioPlayer'
```

t0mpr1c3:
OK, I'll push a bugfix.

Issue #139, opened 2016-11-26
Choose machine type in ReqStart Message


dl1com:
Get rid of necessity to compile different firmware versions for KH930 and KH910, choose version when sending ReqStart Message.

This implies a also a change in the AYAB Firmware and API

chris007de:
We should try to get this into the 1.0.0 firmware.
@Sturla22 has a WIP for the firmware part of this issue: https://github.com/AllYarnsAreBeautiful/ayab-firmware/issues/29

Not sure if my work in https://github.com/AllYarnsAreBeautiful/ayab-desktop/tree/v0.9_mono_firmware is still relevant, perhaps it's worth to simply start work on this issue over again.

To Dos:

- [ ] Update AYAB protocol to be able to set machine type from Host
- [ ] Remove unnecessary UI elements for choosing machine type from Firmware widget
- [ ] Add machine type to preferences menu
- [ ] Add machine type to configuration object

t0mpr1c3:
OK, I'll have a go at the UI changes. It's timely, the firmware flash UI needed a few changes anyway.

t0mpr1c3:
#3 implemented here: https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/373

t0mpr1c3:
Firmware changes implemented here: https://github.com/AllYarnsAreBeautiful/ayab-firmware/pull/34
AYAB API v6 documented here: https://github.com/AllYarnsAreBeautiful/ayab-manual/pull/13

Issue can be closed once firmware changes have been tested.

t0mpr1c3:
In the current version of the API, machine type is specified in the `reqInit` message not `reqStart`.

dl1com:
Implementation is done, closing.

PR #386, opened 2020-07-14
Refactored firmware flash UI without hardware options.

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/139

Changed the form so that there are no hardware options to select. Also changed the behaviour so that the `Flash` button is not enabled unless a port has been selected, and fixed the system call to `which avrdude`.

Some other minor changes, including changing the function that updates the notification field and the signal that connects to it.

The firmware `.hex` files that the `firmware.json` file points at are dummy files only.


chris007de:
Do you think we might ever have to need again to propose multiple firmware versions in the flash UI? Otherwise, we could get rid of the bottom selection window as well and just print the name of the relevant *.hex file which was shipped with the Application.

t0mpr1c3:
I'm not sure, to be honest. Maybe.

I could imagine a situation where we want to release a patch for the firmware without a new desktop release. We'd probably want to have the user put the patched `.hex` files to the appropriate folder under `resources` and have the desktop software automatically discover it.

At the moment the list of available files is coded as a `.json` file, but it would be easy enough to do automatic file discovery as long as there is a consistent schema for naming the `.hex` files.

Issue #375, opened 2020-07-09
Set up of application environment


dl1com:
(Triggered by #371)

- To be able to run the software after cloning the repository and to properly generate frozen packages for Windows and macOS, some prerequisites have to be fulfilled.

  - [x] Generate the translation files. This can already be done with the `ayab_trans.pl` script.

  - [x] To be stringent, we may also require the Python ui files generated from the .ui files to be generated (e.g. ayab_about_gui.py from ayab_about_gui.ui), just as we do for the .qm files. As we already left the comfort zone of "clone the repo and run the software" by requiring the translation files to be generrated, this would not yield additional effort. A script is required to run `pyuic` on the .ui files. (to be discussed, @t0mpr1c3 ?)

  - [x]  We should download the avrdude binaries for Windows and macOS automatically from https://download.savannah.gnu.org/releases/avrdude/, unpack them and put them into the right place. (to be discussed, @t0mpr1c3 ?)

The three scripts for these actions could be called by a script (e.g. "setup_environment.sh").
Running this script after cloning the repository sets up a proper environment for running the application using `fbs run`.
Running this script would also be the foundation for running the existing build scripts (=generate Windows Setup and macOS application)
(for Linux we still have to finally decide how to distribute the software to non-developers/non-git users...tar.gz, pypi, AppImage,...?).

t0mpr1c3:
Sounds good, Chris.

I don't have strong feelings about how to distribute the Linux. I guess I would go with AppImage, if it integrates easily into our pipeline? I would guess that tarballs might need to be versioned for different distros, and Pypi isn't going to reach many people who don't already use Git.

t0mpr1c3:
I've written a script to generate Python files using `pyuic5`. It relies on the filenames having already been renamed, though, so I'll wait until after the renaming pull request is resolved before requesting the changes.

t0mpr1c3:
I'm having a go at making a `build.yml` for the AppImage, which I'll put in branch `1.0.0-appimage`.

t0mpr1c3:
This is what I've found out so far about building Linux apps:

1. `fbs` supports Ubuntu/Debian, Fedora, and Arch distros providing Docker images in which the packages can be built.

2. Github workflows provide an Ubuntu runner in which the `.deb` package can be built. We would have to provide self-hosted runners to build Fedora and Arch packages.

[...]

Issue #310, opened 2020-06-08
Set up Github actions for building AYAB


chris007de:
- [x] Linux
    Github Runner: 18.04, 20.04
- [x] Windows
    Github Runner: Windows Server 2019
- [x] macOS
    Github Runner: macOS 10.15

https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

Notes:
- We have to check if build under Windows Server 2019 is executable under Windows 10 properly.
- If we build under macOS 10.15, the system we will run the software has to be at least 10.15 also. Therefore, we might want to set up a custom runner with an older version of macOS for better compatibility with older devices.

t0mpr1c3:
I really don't know a whole lot about this, but if you give me some pointers I could start on the Linux.

chris007de:
In general, Github actions work a lot like Travis CI, Appveyor etc.
There is one more more a "workflow" files which describes what happens when certain actions occur in the repository.

Here you can find our workflow file for the automatic run of the pytests on each PR or push to a branch: https://github.com/AllYarnsAreBeautiful/ayab-desktop/actions/runs/138826515/workflow
So basically the yml file describes the triggers and what happens subsequently. It is stored in the repo under .github/workflows

Here, I began setting up the action for the Windows build:
https://github.com/AllYarnsAreBeautiful/ayab-desktop/actions/runs/137632800/workflow

For each OS, the basic flow is the same:
- checkout the code
- install prerequisites
- run `python -m fbs freeze` to build binary
- build installer with `python -m fbs installer` (Builds a Setup for Windows, the App for macOS, to be skipped for linux)
- deploy artifacts
- If commit was tagged: create a github release and attach the artifacts to the release
  - bonus: upload to pypi

For the 0.95 release, we used Gitlab pipelines with self-hosted Gitlab runners (virtual machines on my computer) to do this (see gitlab.yml and the build scripts in the linux-build, macos-build and windows-build folders) in the root folder), but as Github added actions as a feature in the mean time, we may switch over to Github actions. So we would be no longer depending on my Gitlab/Virtual Machine runner setup :)

t0mpr1c3:
I really don't know much devops, but the documentation looks pretty good. I would imagine Windows is the highest priority. Although I do I have a Windows 10 box I haven't attempted a build yet.

Off topic but I think the GUI looks better in Ubuntu than Windows.

chris007de:
The Windows build works.

[...]

PR #391, opened 2020-07-17
Added github workflows to build Ubuntu/Debian and Mac OSX packages.

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/375, https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/310

The Fedora and Arch workflows require GPG keys to build the Docker VMs, ~~which I guess need to be written to the `secrets` settings in the repo. I don't have the permissions to do that.~~ I found a way to write the GPG password and build the VMs on the github runner using `fbs buildvm`, but running them with `fbs runvm` failed. I succeeded in building Arch and Fedora packages locally, though.


chris007de:
Thanks a lot for your efforts! I'll need some time to work through this and will give you feedback.
But what I can already tell from skipping through the changes: We'll probably not wanna store the private key for signing the binaries publicly available in our repository :)
I'll check how to deposit such a key as a github secret.

.github/workflows/build-fedora.yml
```diff
@@ -0,0 +1,64 @@
...
+      run: echo "::set-env name=GIT_TAG::$(git describe --tags)"
+    - name: Set PACKAGE_VERSION
+      run: |
+        echo ${{env.GIT_TAG}} > src/main/resources/base/ayab/package_version
+        sed -i 's/PACKAGE_VERSION/${{env.GIT_TAG}}/' src/build/settings/base.json
+        cat src/main/resources/base/ayab/package_version
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
```
chris007de, line 31: Perhaps make also use of `strategy: matrix: python-version...`? -> ${{ matrix.python-version }}

.github/workflows/build-ubuntu.yml
```diff
@@ -0,0 +1,54 @@
...
+        echo "::set-env name=GIT_TAG::$(git describe --tags)"
+    - name: Set PACKAGE_VERSION
+      run: |
+        echo ${{env.GIT_TAG}} > src/main/resources/base/ayab/package_version
+        sed -i 's/PACKAGE_VERSION/${{env.GIT_TAG}}/' src/build/settings/base.json
+        cat src/main/resources/base/ayab/package_version
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
```
chris007de, line 32: see above, define python-version in matrix and make use of ${{ matrix.python-version }}

.github/workflows/build-arch.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      # - '*-dev'
+
+jobs:
+  build:
+    runs-on: ubuntu-16.04
```
chris007de, line 14: Should be running with ubuntu-latest as well, shouldn't it?

.github/workflows/build-arch.yml
```diff
@@ -0,0 +1,67 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Use old pip version
+      run: python3 -m pip install pip==18.1
+    - name: Update setuptools
+      run: python3 -m pip install --upgrade setuptools
+    - name: Install fbs
+      run: python3 -m pip install fbs==0.8.6
+    - name: Create Docker image
+      run: |
+        echo '{"gpg_pass": "password"}' > src/build/settings/secret.json
```
chris007de, line 42: I can store the gpg password as secred in the github repo (see https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets-for-a-repository), please send it to me over a secure channel.

.github/workflows/build-fedora.yml
t0mpr1c3: OK

.github/workflows/build-ubuntu.yml
t0mpr1c3: OK

.github/workflows/build-arch.yml
t0mpr1c3: For the time being I am keeping the Linux scripts on Ubuntu 16.04 for the sake of compatibility as this is still a supported version. Anything that works on 16.04 should also work on 18.04, but the reverse is not necessarily true.

.github/workflows/build-arch.yml
t0mpr1c3: This is a dummy password anyway! We can do this when the scripts are more mature.

t0mpr1c3:
> * In production, I'd propose to run the workflows only on when a tag is pushed (see https://github.com/trappitsch/fbs-release-github-actions/blob/master/workflows/release-ubuntu.yml#L5). Generating installers (and Github releases) is not necessary for each and every push in my opinion.

Yes.

>   For now, during development, you could define to run only on push to branch `1.0.0-build` so we see how everything behaves.

OK.

> * Please add the steps for creating releases and upload the results (just as shown in the https://github.com/trappitsch/fbs-release-github-actions examples)

Will do.

> * If we switch our changelog format to https://keepachangelog.com/, we could even make use of https://github.com/marketplace/actions/changelog-reader later ;)
> * We can remove ayab_logo_rc.py now, as it is generated automatically, right?

Yes.

t0mpr1c3:
In the latest commit have removed `build-arch.yml` and `build-fedora.yml` as works-in-progress. I now have a script to build an AppImage that compiles everything except the audio packages (which are not packaged as wheels -- the `setup.py` scripts break). So we can have a silent cross-Linux version.

linux-build/appimage/ayab.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=ayab
+Exec=ayab
+Comment=All Yarns Are Beautiful
+Icon=ayab
+Categories=Graphics;
+Terminal=false
+X-AppImage-Version=v0.95-104-gc97f0b6
```
chris007de, line 9: I guess we should update this field dynamically, as we do with `package_version`...time for another script? :D

requirements.txt
```diff
@@ -1,14 +1,15 @@
-pip==18.1
+#pip==18.1
```
chris007de, line 1: Is pip==18.1 no more hard requirement?

linux-build/appimage/ayab.desktop
t0mpr1c3: I think this field is updated automatically.

requirements.txt
t0mpr1c3: Correct.

t0mpr1c3:
I put the Ubuntu, Mac, and Windows release scripts in a single file. It's definitely better this way, because they release assets can all be attached to the same release tag. I haven't tested the Mac `.dmg`, but the other two seem to work. 

I think we can close this PR. If I can get the AppImage running I'll raise a separate PR.

Issue #278, opened 2018-12-28
Create Linux AppImage for deployment


chris007de:
For instructions see:
https://github.com/AppImage/docs.appimage.org/blob/master/source/packaging-guide/manual.rst

PR #394, opened 2020-07-22
Added AppImage asset to release workflow.

t0mpr1c3:
Tested very briefly on Ubuntu 16.04. The sound works, don't know about `avrdude`. Should really test it on a live CD of Fedora or something.

To run the AppImage, just `chmod +x` and double click. Unfortunately the nifty icon does not show unless you install a daemon :/

In theory it should be straightforward to make the AppImage updatable from a menu item (https://github.com/AppImage/QtAppImageUpdateExample).

Completes issues https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/278 and https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/310. Once `avrdude` is tested this may also close https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/375.


t0mpr1c3:
> * I did not test the AppImage yet...how is the storage of user-preferences and pattern handled for AppImage (is it still allowed to interact with the home directory of the user or is it completely isolated?)

Preferences use `QSettings`, which for Linux saves a file `~/.config/AYAB/ayab.conf`. This remains the case for the AppImage.

> * If we only build on push of tags, we could remove the recursive clone of the git repos in the jobs, as we are sure that we have a tag available for the git --describe commands in this case.

It does feel redundant and I understand that there is a different method of obtaining tags. But I am still rather inexperienced with these workflows and I don't know what else would need to change in order to not have to clone the repos. Feel free to edit the workflows if you know what to do!

> * Additionally to the upload of the build results to a release, we could upload the binaries as an artifact of the job (https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#uploading-build-and-test-artifacts), so we have them available for testing, even if we delete a release which as just created for testing.

This was an easy addition, implemented in my most recent commit.

t0mpr1c3:
By the way, I think the oudated build check ("Expected - Waiting for status to be reported") must be edited in the repository settings, which I do not have access to ;)

probonopd:
Where can it be downloaded from?

t0mpr1c3:
@probonopd Hi there, and thanks for your work developing the AppImage software! At the moment there is no release version of the AYAB AppImage. That will change with the 1.0 release which could be another couple of months. I'll raise a documentation issue to link to it.

probonopd:
Thank you very much @t0mpr1c3

Issue #133, opened 2016-11-26
 Integrate KH-270 support to GUI


chris007de:
The KH-270 does only have 114 needles, so the GUI should offer a mode to take this significant difference into account. This mainly concerns pattern display/handling and start/stop needle ranges.

PR #396, opened 2020-07-23
Allow variable machine width (required to support KH270).

t0mpr1c3:
Referencing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/133.

Relevant graphical methods and tests updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,33 +8,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 1.0.0 / Unreleased
 
+* #395 GUI: Check for new release on startup
+* #393 Documentation: Add forthcoming features to CHANGELOG
 * #389 GUI: Rollback to Python3.6, PyQt5==5.9.2
-* #379 GUI: Fix bug in menu bar
-* #371 Documentation: Add installation of avrdude for Linux
+* #379 GUI: Change appearance of menu bar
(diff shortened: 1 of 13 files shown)


t0mpr1c3:
I think I pushed the wrong version. It should work now, can you try again?

chris007de:
> I think I pushed the wrong version. It should work now, can you try again?

Still not working. I think setting the KnitMode enum entries to `auto()` broke it. When changing it back to the static assignments, the program starts properly again.

Also:
- Closing the window does not quit the application properly, the log just shows `QXcbConnection: XCB error: 3 (BadWindow), sequence: 806, resource id: 10507589, major code: 40 (TranslateCoords), minor code: 0` and the application has to be closed with ctrl+c
- The combobox for languagues is empty in the preferences menu. I ran `setup_environment.sh` prior.

t0mpr1c3:
OK, I'll try reinstalling my development environment. I'm a bit confused that I am not seeing the same errors.

t0mpr1c3:
OK, I have fixed the language bug and the auto() bug. I can't replicated the problem on closing the window.

chris007de:
> OK, I have fixed the language bug and the `auto()` bug. I have not replicated the problem on closing the window.

Ok, language and KnitMode is working for me now.
Exiting the application still fails, but let's handle this in another issue (in case it happens for somebody else than me, also)

Issue #370, opened 2020-07-06
Need translations for the firmware flashing utility


t0mpr1c3:
Just about everything in the software has been translated except this. I can fill in the title labels, but ideally we would have translations for the guide as well.

t0mpr1c3:
The way the firmware is made available will change (https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/414), so translation can wait until after those changes.

dl1com:
User djdollar on Discord also reported missing translations (did not check against the list above)

![Image](https://github.com/user-attachments/assets/85766c96-c71a-48a5-b4c3-cbd944e042b8)

![Image](https://github.com/user-attachments/assets/25fc5289-7cf2-4ba2-8034-0217bebae3cd)

![Image](https://github.com/user-attachments/assets/3affcbba-ba72-42f5-8626-07b6d0293c2b)

![Image](https://github.com/user-attachments/assets/fd7bbb10-5884-43c4-9db2-989be8c5c590)

![Image](https://github.com/user-attachments/assets/22481baa-5338-4290-b435-e1a7bd0bc5fb)

![Image](https://github.com/user-attachments/assets/87769627-e1eb-41c5-b255-7f128a821521)

plus Russian translations for these fields

```
Zoom in Приблизить
Zoom out Отдалить
To Be Selected Иглы к выбору
Knitting canceled Вязание завершено
Machine Машина
Default Knit Side Image  Отражать рисунок по умолчанию
Aspect Ratio Отношение сторон
Disable Hardware Beep Отключить встроенный динамик
Lower Display Stitch Width Ширина стежка в нижнем окне 
```

dl1com:
## Current status
Mostly fixed with #766 and #767 

## Still missing translations (reported by djdollar):

![Image](https://github.com/user-attachments/assets/efe2301d-e96b-4ced-b4a2-65a20ce45a11)

![Image](https://github.com/user-attachments/assets/34c81f97-df35-4591-a1c0-16e9035c74f7)

![Image](https://github.com/user-attachments/assets/b33d382d-54ff-435a-9309-6685acb7da93)
(Help - About -> "Помощь - о программе")
 https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/main/src/main/resources/base/ayab/translations/ayab-translation-master.tsv#L55)

![Image](https://github.com/user-attachments/assets/01e66631-308d-488b-9b18-c42b0cfcd678)
[...]

Issue #206, opened 2017-06-25
No error if you flash the wrong firmware


Adrienne200:
Flash the 930 firmware, then use it on a 910, there should be an error. Or are we working towards a single firmware, or being OK with either 0.8 firmware?

On a 910, flashed with 930etc 0.8 firmware, 0.9 software:
Set up to knit a simple pattern, default settings. 
Expected: It should error when you press Knit.
What’s actually happening is that it half-works; it selects row 1 correctly and beeps, fails to select row 2 and no beep even if I cross the turn mark, then on the third pass it selects row 2 and beeps. Basically mostly only working on the left-right passes. Only sometimes beeps on the left.
Tried a current EMSL board and a Tinkersoup original, same.
Log indicates that the data is being sent down correctly.
Looks like it did this in 0.8 too on this machine and I hadn’t noticed. 0.75 does give the error that I expected; if it showed now it should have a wording change, #127 

The reverse case, I'm not sure. 910 firmware on a 930 machine also doesn't give the error, but it appears to work as usual, at least in a quick basic test. 
Then one more test, 930 fw on a different 910 machine. First test worked, second time through it failed as above.

[ayab_log 830 fw on 910.txt](https://github.com/AllYarnsAreBeautiful/ayab-desktop/files/1100548/ayab_log.830.fw.on.910.txt)

oskay:
It is probably a good idea to query the firmware version and configuration. I'd recommend that we target this as a milestone post 0.9.

chris007de:
@oskay we are already querying the firmware version at the very beginning of the serial communication. If the firmware version is not as expected, the GUI shows an error.
Also, I would probably remove any non-compatible firmware images from the release, so people cannot even flash images which are incompatible with the GUI they are currently running.

@Adrienne200 
We were already working on some kind of "monolithic" firmware which covers all machine models (https://github.com/allyarnsarebeautiful/ayab-desktop/tree/v0.9_mono_firmware), but I put it on hold as we started doing serious changes (python3, pyqt5).
The goal was to remove the need to decide for a machine model when flashing the Arduino, so we only have a single firmware hex file to distribute (of course, there would be still separate images for different Arduino models such as Uno, Mega).
[...]

Issue #200, opened 2017-06-17
Hang if you try to flash the firmware with no Arduino attached.


Adrienne200:
0.9 RC4
Mac or Windows

Launch the AYAB software, but do not plug in the USB to the hardware.

Tools menu - Load AYAB firmware
select anything: 910, Uno, 0.8
Click Flash.
Nothing is happening, but you can't close the firmware window.

Expected, either:
 - the Flash button should not be live if it hasn't found an Arduino
 - it should check the port first and show an error message.
I prefer the second one.

oskay:
Confirmed (Mac).

chris007de:
So I guess we want to have some kind of Arduino detection.
Here are some examples:
- https://stackoverflow.com/questions/24214643/python-to-automatically-select-serial-ports-for-arduino
- https://arduino-er.blogspot.de/2015/04/python-auto-detect-arduino-connect.html

Let's see how reliable that works, we should avoid to lock out users with exotic Arduino clones.

oskay:
Without adding a library, would it be possible to simply detect that the selected serial port is "sensible" before attempting to flash?

chris007de:
We are already importing "serial" to do the actual communication, so this would not be a big deal.
Detection based on sensible portnames would a very basic approach, but might also yield false positives (e.g. checking for /dev/ttyACM* or /dev/ttyUSB*).
I'll have a look into it and propose a solution.

chris007de:
Still occurs. If I try `Flash` on a serial port where no Arduino is connected, avrdude is just stuck forever.

t0mpr1c3:
Since we poll the subprocess for output, we should be able to signal a fail if the process times out. 10 seconds should be enough, right?

chris007de:
> Since we poll the subprocess for output, we should be able to signal a fail if the process times out. 10 seconds should be enough, right?

Yes, 10 secs should be a reasonable timeout.

dl1com:
Trying to flash the firmware without having an arduino attached shows error box "Error flashing firmware".
Tested with test230514 (debian).

Closing...

PR #406, opened 2020-07-26
Translations and 10 second timeout for firmware flash GUI.

t0mpr1c3:
Partial translations for firmware flash GUI (https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/370). Plus Korean language translations although these still need font support.

Added 10 second timeout for firmware flash `avrdude` subprocess (https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/206, #200).

Refresh USB port combobox each time that the dialog is opened.


src/main/python/ayab/firmware_flash.py
```diff
@@ -147,32 +150,17 @@ def execute_flash_command(self):
         # else
         tr_ = QCoreApplication.translate
         try:
-            p = Popen(command, stdout=PIPE, stderr=STDOUT, shell=True)
```
t0mpr1c3, line 150: On second thoughts I'm not satisfied with this implementation of timeout because it hands over control of the thread to the shell process and freezes the GUI.

chris007de: Do you intend to change this or is this PR ready to merge?

t0mpr1c3: Let's merge the PR and put a "please test" label on the issue. I'll get back to it at a later date.

PR #411, opened 2020-07-28
Changes from API v5 to API v6 and code annotations to help ensure backwards compatibility after API changes.

t0mpr1c3:
Added parameter to `req_start` method so that it specifies the machine in use. Also added CRC8 checksum.

Added device state parameter to method in `Status` class that parses carriage info so that error states can be evaluated.

Added color byte to `cnf_line` message and shortened needle data array to 15 bytes for KH-270 machines.

Incremented (minimum) required API version to 6.

Added annotations and renamed a bunch of methods with `_API6` suffix that are (potentially) specific to this API version.

Changes documented here: https://github.com/AllYarnsAreBeautiful/ayab-manual/pull/13/files#diff-96cc378dcdef180641176501f42183c7

### DIFF
--- a/src/main/python/ayab/engine/communication.py
+++ b/src/main/python/ayab/engine/communication.py
@@ -35,10 +35,14 @@ import pprint
 class MessageToken(Enum):
     unknown = -2
     none = -1
-    cnfStart = 0xC1
+    reqInfo = 0x03
     cnfInfo = 0xC3
-    reqLine = 0x82
+    reqTest = 0x04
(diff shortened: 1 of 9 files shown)


chris007de:
@t0mpr1c3 are you still working on this PR or is it ready to merge?

t0mpr1c3:
It's ready to go.

t0mpr1c3:
Subsumed into new PR #412.

PR #430, opened 2021-09-30
.cut file support

t0mpr1c3:
Allows patterns in .cut format to be uploaded as greyscale images. (In this format, the colour palette is specified in a separate file with a .pal suffix. I have not implemented a way to load in these files.)

This feature was suggested by Alessandrina Costa as potentially useful for Passap users who design patterns using Wincrea.


chris007de (approved):
:clap:

PR #435, opened 2022-05-30
Fix a few bugs with the status panel display

clholgat:
The logic to hide the color column was hiding the carriage/direction column and the UI. Updated this logic to deal with a variable number of columns.

Every time the status checker looped, it would add a new row to the widget whether it needed to or not. Changed this so that the UI only updates if any of the status that contributes to the UI has changed.


chris007de (approved):

chris007de:
Very sensible change, thank you!

PR #436, opened 2022-05-30
Fix inverted needle selection for singlebed

clholgat:
This is supposed to be on top of https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/435 but github doesn't do a real diffbase. It can wait until that one goes in.

In singlebed mode, we need to select the needles for color 1 not color 0. Color 0 is the background. Color 1 is going to be whatever's in the second yarn feeder.

### DIFF
--- a/src/main/python/ayab/engine/communication.py
+++ b/src/main/python/ayab/engine/communication.py
@@ -169,7 +169,7 @@ class Communication(object):
         # fallthrough
         self.__logger.debug("unknown message: ")  # drop crlf
         pp = pprint.PrettyPrinter(indent=4)
-        pp.pprint(msg.decode())
+        pp.pprint(msg[1: -1].decode())
         return msg, Token.unknown, 0
 
     def read_API6(self):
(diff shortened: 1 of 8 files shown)


chris007de:
I am not too sure about the reason for the change - @clholgat could you please elaborate?
Is it just to consolidate the color handling (0 -> always background, > 0 -> colors) or was the behaviour not as expected (inverted colors?)?

clholgat:
Sorry for the late reply, github isn't emailing me about reviews.

The behavior wasn't as expected for single bed, two color fair isle. Color 0 is white, Color 1 is black. So when knitting, the white pixels would translate to selected needles creating inverted colors.

chris007de:
@clholgat sorry for leaving this open so long - is it ready to merge?

clholgat:
Abandoning because I think the problem is elsewhere: https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/442

PR #464, opened 2023-05-10
Fix bug in Windows installer

t0mpr1c3:
This bugfix involves downloading a patched version of fbs from my personal github account. It's probably more sensible to host it under AYAB itself, or incorporate it as a submodule.

Going forward there are likely going to be more issues requiring patches to fbs as the free version gets increasingly out of date.

Edit: Note that we could get the desired fix by bumping the fbs version from 0.8.6 to 1.0.1, but that might cause problems with the way that we name versions at the moment: after version 1.0.0, fbs insists that all version names are formatted as `r'\d+\.\d+\.\d+$'`


t0mpr1c3:
The bleeding edge release is https://github.com/t0mpr1c3/ayab-desktop/releases under the tag testa579a88

.github/workflows/build-multi-os.yml
```diff
@@ -272,6 +272,11 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Patch requirements.txt to use patched version of fbs
+      shell: bash
+      run: |
+        cat requirements.txt | sed 's/fbs==0.8.6/-e git+https:\/\/github.com\/t0mpr1c3\/fbs.git@fbe63505acdf5475e914930ea9de958384a5128c#egg=fbs/' > patched-requirements.txt
```
dl1com, line 278: @t0mpr1c3 I would like it better to persistently change the requirements.txt rather than patching it in the workflow, as it somehow "hides" the change.
I'm OK with pointing to the fork at your account, but we can also store it in a repo under the AYAB orga, whatever you prefer.

t0mpr1c3: @dl1com I would prefer it be under AYAB so that other people have access. Please could you make an AYAB org fork of fbs, and then I'll raise a PR.

t0mpr1c3: > I would like it better to persistently change the requirements.txt rather than patching it in the workflow, as it somehow "hides" the change. 

Yes I agree.

t0mpr1c3:
The linter doesn't like some python syntax in the patched fbs repo. This is code that should already be there under v0.8.6, and dates back a couple of years, so I don't think this is a Python version issue. Any ideas?

.github/workflows/build-multi-os.yml
dl1com: > @dl1com I would prefer it be under AYAB so that other people have access. Please could you make an AYAB org fork of fbs, and then I'll raise a PR.

Here we go: https://github.com/AllYarnsAreBeautiful/fbs

dl1com:
> The linter doesn't like some python syntax in the patched fbs repo. This is code that should already be there under v0.8.6, and dates back a couple of years, so I don't think this is a Python version issue. Any ideas?

That's strange behavior...I'd say it _should_ work.
One reason could be that the fbs version 0.8.6 pulled from pypi by our requirements.txt is not exactly the same as the version tag 0.8.6 in the fbs repo. Don't know if it's worth to do a diff?
Does flake8 pass on your local system?

t0mpr1c3:
diff indicated no differences at all for the file that flake puked on.

I think flake is testing the fbs code in a way that would not otherwise be done.

t0mpr1c3:
That fixed the linter, now pytest has found an undeclared dependency.

t0mpr1c3:
More pytest errors, so I'll try exempting fbs from pytest altogether.

t0mpr1c3:
More pytest errors. Patching the patch isn't working for me so I am going to just apply a patch file during the build. This means the windows build.sh will need to be updated as well.

t0mpr1c3:
This now works as far as it goes. The only remaining issue that I know of is that the windows `build.sh` script needs to be changed. My preferred way of installing would be `conda env create` and then patching the source file which should be under $Env:HOMEPATH\Anaconda3. Any thoughts on this?

t0mpr1c3:
Modified build script in README tested working on Windows 11.

`build.sh` files deleted because the releases are created by github actions now.

t0mpr1c3:
README.md install script for Linux tested good in Ubuntu 20.04. Mac not tested

dl1com (approved):

PR #468, opened 2023-05-11
Download firmware matching manifest and incorporate into builds

t0mpr1c3:


.github/workflows/build-multi-os.yml
```diff
@@ -37,6 +37,129 @@ jobs:
         echo "sha-short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
         echo "tag=$(git describe --tags)" >> $GITHUB_OUTPUT
         echo "draft=$(git describe --tags | sed -e 's/^test.*/true/;s/^v.*/false/')" >> $GITHUB_OUTPUT
+    - name: Get firmware version matching manifest
+      run: |
+        manifest=$(cat src/main/resources/base/ayab/firmware/manifest.txt)"
+        available=$(curl -s https://github.com/AllYarnsAreBeautiful/ayab-firmware/releases | grep 'releases/tag/v' | sed 's/.*releases\/tag\/v//;s/\".*//')"
```
dl1com, line 43: Am I interpreting this correctly and this works only for releases named "v..."?
I guess it would be good if the releases named "test..." would be supported as well for now so we can generate test/beta releases.

t0mpr1c3: Yeah let's do that.

dl1com (approved):

PR #470, opened 2023-05-14
Fix conda command for setting up 3.6 environment

dl1com:
Calling it with 3.6.* fails on my debian system. Calling it with 3.6 should use the latest 3.6.x version as well.


t0mpr1c3 (approved):
Looks good, Chris.

Issue #461, opened 2023-05-01
Software state broken after "Test AYAB Device"


VIPQualityPost:
branch: 1.0.0-dev
os: macOS

Without any Arduino connected, if you open the hardware test panel, then close it, if you load an image you will be unable to use the controls on the side panel (all grey, clicking even while grey does not have any effect), the only way to restore operation to the software is closing and re-opening. I don't know if this happens with a device actually connected.

t0mpr1c3:
Which release/commit were you using?

t0mpr1c3:
This might be a bug in the FSM and/or API. Presumably the desired behaviour is that if contact with the device is lost or not established then after a period of time an error code should be emitted and the FSM should revert to the initial, disconnected state. This should also happen if the engine or test module closes without warning.

t0mpr1c3:
I suggest a 10 second timeout as discussed here: https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/200

dl1com:
Build: https://github.com/AllYarnsAreBeautiful/ayab-desktop/releases/tag/test230515 in Windows 10

After starting the AYAB application and directly opening the Hardware Test without any hardware attached and Port set to "simulation", the Hardware Test Windows shows up but lags and is unresponsive. Only possibility to fix is to close the AYAB application completely.

Log shows:
```
2023-05-17 21:58:52,380 root     DEBUG    Entered state NO_IMAGE
2023-05-17 21:58:58,158 root     DEBUG    Entered state TESTING_NO_IMAGE
2023-05-17 21:58:58,159 Control  DEBUG    State CONNECT
2023-05-17 21:58:58,159 Control  DEBUG    Port name: Simulation
2023-05-17 21:58:58,159 Control  DEBUG    State VERSION_CHECK
2023-05-17 21:58:58,159 Control  DEBUG    State VERSION_CHECK
2023-05-17 21:58:58,160 Control  INFO     API v6, FW v1.0
2023-05-17 21:58:58,160 Control  DEBUG    State REQUEST_TEST
2023-05-17 21:58:58,160 Control  DEBUG    State CONFIRM_TEST
```

t0mpr1c3:
In [test230518](https://github.com/AllYarnsAreBeautiful/ayab-desktop/releases/tag/test230518) on Windows 11, the FSM gets stuck in state REQUEST_TEST and the dialog never loads.

t0mpr1c3:
[...]

PR #478, opened 2023-05-14
Fix persistence of test thread after test dialog is closed

t0mpr1c3:
Addressing https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/461, https://github.com/AllYarnsAreBeautiful/ayab-firmware/issues/62


src/main/python/ayab/pattern_import.py
dl1com, line 1: pattern_import included in the PR by mistake I guess? :)

t0mpr1c3: oops

t0mpr1c3: The problem is that pattern_import.py has CRLF line endings in the repo, and my git keeps trying to change them to LF.

dl1com (approved):

PR #485, opened 2023-05-18
Remove an extra machine param from req_start

clholgat:
Firmware was crashing after the latest FSM updates to set machine type before doing carriage detection. 

Turns out I missed one param in req_start.


X-sam (approved):
LGTM. Nice catch, that's a really annoying issue to have run into!!! My kingdom for static typing.....

t0mpr1c3:
Please can you verify that the protocol remains compliant with the API which is documented [here](https://github.com/AllYarnsAreBeautiful/ayab-manual/pull/13).

Issue #506, opened 2023-05-20
audio error after knitting finishes


t0mpr1c3:
on Ubuntu 22.04

```bash
ALSA lib pcm.c:8568:(snd_pcm_recover) underrun occurred
```

on Windows 11

```pwsh
Traceback (most recent call last):
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\main.py", line 112, in <module>
    exit_code = appctxt.run()  # 5. Invoke run()
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\main.py", line 46, in run
    return self.app.exec_()  # 3. End run() with this line
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\ayab\ayab.py", line 153, in finish_operation
    self.audio.play("finish")
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\ayab\audio.py", line 85, in play
    self.__worker.play(sound, blocking=False)
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\ayab\audio.py", line 43, in play
    wave_obj = self.__wave(sound)
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\ayab\audio.py", line 55, in __wave
    self.__cache[sound] = self.__load_wave(sound)
  File "C:\Users\Tom\Documents\hack\ayab-desktop\src\main\python\ayab\audio.py", line 62, in __load_wave
    wave_read = wave.open(path.join(self.__dir, filename), 'rb')
  File "c:\programdata\miniconda3\envs\ayab\lib\wave.py", line 499, in open
    return Wave_read(f)
  File "c:\programdata\miniconda3\envs\ayab\lib\wave.py", line 163, in __init__
    self.initfp(f)
  File "c:\programdata\miniconda3\envs\ayab\lib\wave.py", line 130, in initfp
    raise Error('file does not start with RIFF id')
wave.Error: file does not start with RIFF id
```

There are two problems here: (1) some issue with the WAV file, (2) Windows is not following the symbolic link to the WAV file.

jonathanperret:
Windows and Linux are reported fixed in #514; I can confirm on macOS I don't get an audio error at the end of (simulated) knitting either. Looks like this can be closed?

PR #507, opened 2023-05-20
disable beep after knitting

t0mpr1c3:
temporary fix for https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/506


dl1com:
> temporary fix for #506

Same question as for the other temporary fix - what's the chance to fix the reason for this behavior?
Otherwise we'll end up with a bunch of "temporary" fixes which become final.

t0mpr1c3:
> > temporary fix for #506
> 
> Same question as for the other temporary fix - what's the chance to fix the reason for this behavior? Otherwise we'll end up with a bunch of "temporary" fixes which become final.

I don't intend to chase these down right away, but I can put these temporary fixes in a separate branch if you would prefer.

Issue #537, opened 2023-06-16
Log current statemachine state only on change


dl1com:
Currently, the logfile is cluttered with regular reporting of the current FSM state, which bloats it up and decreases readability.

FSM states shall only be logged when changed.

```
2023-06-16 09:26:19,033 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,139 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,242 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,345 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,451 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,558 Control  DEBUG    State VERSION_CHECK
2023-06-16 09:26:19,664 Control  INFO     API v6, FW v1.0
2023-06-16 09:26:19,666 Control  DEBUG    State INIT
2023-06-16 09:26:19,774 root     INFO     Notification: Initializing firmware
2023-06-16 09:26:19,775 Control  DEBUG    State INIT
2023-06-16 09:26:19,879 root     INFO     Notification: Please start machine. (Set the carriage to mode KC-I or KC-II and move the carriage over the left turn mark).
2023-06-16 09:26:19,881 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:19,989 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,096 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,198 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,302 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,407 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,515 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,623 Control  DEBUG    State REQUEST_START
2023-06-16 09:26:20,730 Control  DEBUG    State REQUEST_START
...

```

t0mpr1c3:
The state machine should not be re-entering the same state repeatedly in the first place.

To be honest, the whole engine state machine could do with being rewritten in a declarative style.

PR #538, opened 2023-06-16
Log current engine FSM state only on change

t0mpr1c3:
Addressing #537

### DIFF
--- a/src/main/python/ayab/engine/control.py
+++ b/src/main/python/ayab/engine/control.py
@@ -47,6 +47,7 @@ class Control(SignalSender):
         super().__init__(parent.signal_receiver)
         self.logger = logging.getLogger(type(self).__name__)
         self.status = engine.status
+        self.notification = Output.NONE
         self.api_version = self.FIRST_SUPPORTED_API_VERSION
 
     def start(self, pattern, options, operation):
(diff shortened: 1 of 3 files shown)


dl1com:
Still pollutes the log with 

```
2023-06-17 11:31:08,482 root     INFO     Notification: Initializing firmware
```

because it repeatedly returns the notification when in state INIT

https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/538/files#diff-14d2314eb34117bd0fd27de70bdfefca9735ba6a4cc85dff28dc10c5fccf2022R127

t0mpr1c3:
I don't see output like that, but I pushed a new commit -- does that fix it?

dl1com:
> I don't see output like that, but I pushed a new commit -- does that fix it?

That's probably caused by the unreliable serial communication I experience on my end.
But thanks for obfuscating the duplicate notifications, now the log looks nice and clear.

dl1com (approved):

dl1com:
Just as a note:
while-looping within the states causes that the UI is non-responsive while being inside that loop. So i.e. waiting for a indState in _API6_request_start renders the cancel button useless.
We should fix that timely.

t0mpr1c3:
That shouldn't happen because the GUI and the engine are in separate threads.

Issue #525, opened 2023-05-31
Only knits one row, then dies.


Adrienne200:
Build: test230530. It selects needles correctly on row 1, then dies as it tries to start row 2.

Tested on:
Build: Desktop test230530, from which you load firmware 230529.
Computer: Mac Intel, 11.7.7 Big Sur
Knitting machine: KH910
AYAB hardware: EMSL interface
(also tested on: Windows 7 and a Shield, same behavior)

Steps:
1. Open any image, I used the sample Triangles_60x10. Single bed, all settings at default. Check that your port selection has found your hardware, is not in Simulation mode.
2. Carriage on KC1 or KC2, cross the left turn mark. Move a bit past it and stop, see #524, and make sure you hear the multiple beep.
3. Continue across. Notice that it has correctly selected needles for row 1, and beeped correctly at the end.
4. Turn around for row 2.
There's a sound from the solenoids, I think they all turned off. Try to continue knitting, all needles are selected to D and there are no more beeps.

clholgat:
The beep delay is probably the extra garter slop that I have in there. I need to cut that down.

I'm seeing incorrect selection from left to right, let me see what's up with that.

clholgat:
My incorrect selection was because I had the wrong machine type selected.

I can't repro the death on row 2 on my 910 from a mac with the latest everything.

zelatyna:
I encountered the same behaviour today. After one row of knitting in fair isle, in the 2nd row all the needles are knit in color A and end in position D :/ 

Also noticed that the bar indicating where the knitting ended is off and shows almost the end of the image
![image](https://github.com/AllYarnsAreBeautiful/ayab-desktop/assets/14292230/aac4a8d8-a657-4c46-a10a-6e3984a28d84)


My specs:
Build: local build from current 1.0-dev branch,  firmware 230529.
Computer: Mac M1, 13.4 Ventura 
Knitting machine: KH950

dl1com:
Just tested in simulation mode of the current 1.0-dev branch with the various test patterns, could not replicate.

@zelatyna could you please share your pattern, just in case it is related to the image?
Plus, please share the logfile, it might be interesting which line numbers are requested/sent, as it seems that the software skips to the end of the pattern.

Adrienne200:
Simulation already worked, this issue only affects actual knitting. 
[...]

PR #540, opened 2023-06-17
Check whether requested line number is in sequence

t0mpr1c3:
If the line number requested by the device is out of sequence, knitting will terminate with an error message.

Helps deal with the problem identified in #525, even though the underlying problem is in the firmware.

### DIFF
--- a/src/main/python/ayab/engine/control.py
+++ b/src/main/python/ayab/engine/control.py
@@ -143,6 +143,10 @@ class Control(SignalSender):
         if self.former_request == self.BLOCK_LENGTH - 1 and line_number == 0:
             # wrap to next block of lines
             self.line_block += 1
+        # requested line number should either be the same as the previous request, or the next line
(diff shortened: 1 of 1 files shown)


dl1com:
I think this will trigger as well in case the firmware re-requests a line number in case of a failed CRC8.
So perhaps also allow line_number == former_request?

(Side note: back in the days, we explicitly allowed the requests of arbitrary line numbers to enable knitting of non-deterministic patterns, i.e. influenced by sensors attached to the AYAB shield or whatsoever. But I think this is a niche-application, so I am fine with this PR :) )

t0mpr1c3:
> I think this will trigger as well in case the firmware re-requests a line number in case of a failed CRC8. So perhaps also allow line_number == former_request?

That's right. The code allows this, and line_number == former_request + 1.

dl1com (approved):

PR #549, opened 2023-07-04
Add instructions about dropping in compiled firmware on linux

Tuuleh:
Helps make testing a bit clearer for Linux users.


README.md
```diff
@@ -183,3 +183,6 @@ The tag can be pushed from your local environment, or via the ["Draft a new Rele
 ### Choosing the firmware release to be bundled with the build
 
 The CI automatically downloads a given firmware release from the [ayab-firmware repo](https://github.com/AllYarnsAreBeautiful/ayab-firmware) and packs it into the Desktop release. The name of the firmware release is chosen in [this manifest file](https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/1.0.0-dev/src/main/resources/base/ayab/firmware/manifest.txt) in the ayab-desktop repo.
+    http://download.qt.io/official_releases/online_installers/qt-unified-mac-x64-online.dmg
```
dl1com, line 186: seems that this line does not belong here?

TEST_GUIDE.md
```diff
@@ -6,9 +6,12 @@ The full list of builds is on the [releases](https://github.com/AllYarnsAreBeaut
 
 If you're just interested in poking around, choose the most recent build.
 
-When you click on a build you should see a list of files under the **Assets** section. Click on the correct file for your operating system. If you're on a mac, you want the **.dmg**, if you're on windows you want the **.exe**, if you're on Linux you'll have to build from source.
+When you click on a build you should see a list of files under the **Assets** section. Click on the correct file for your operating system. If you're on a mac, you want the **.dmg**, if you're on windows you want the **.exe**. Open the file you downloaded and install it the way you normally would. If you're on Linux you'll have to build from source.
+
+
+### Extra Linux setup
```
dl1com, line 12: I think it's an always an issue if you build from source, no matter on which OS

TEST_GUIDE.md
```diff
@@ -30,7 +33,7 @@ Click **Ok** to dismiss the dialog.
 
 ## Install the firmware
 
-These releases have the corresponding firmware bundled in. You no longer need to specify which machine  you have for the firmware.
+These releases have the corresponding firmware bundled in. You no longer need to specify which machine  you have for the firmware. This is with the exception of Linux, where you need to compile and drop in the firmware to upload it through AYAB Desktop (or upload it through other means - described above in the Extra Linux setup section).
```
dl1com, line 36: All releases (also the AppImage for Linux) should contain a firmware file (if this is currently not the case, then it's a bug :( )
The need to put a firmware file there manually is only when building from source, no matter which OS (see comment above)

Tuuleh: Yeah the tar.gz in the release does not contain the firmware hex file - was asked on Discord to just add this to the instructions to make testing a bit easier for whoever works on this next.

TEST_GUIDE.md
Tuuleh: Yeah it's an issue whenever you build from source, independent on OS, but for testing, building from source is only needed for Linux (Windows gets an .exe and Mac gets a .dmg).

README.md
Tuuleh: ups :3

TEST_GUIDE.md
Tuuleh: Or to be honest, running the AppImage failed for me with `./linux-build/appimage/AppRun: line 14: opt/python3.6/bin/fbs: No such file or directory` so I opted to run it with `fbs run` with the hex file dropped into the right location.

dl1com (approved):

Issue #534, opened 2023-06-15
Establish semantic versioning


dl1com:
To be able to fulfil https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/476, we should move to semantic versioning already before releasing 1.0.0, so we can already test the latest version of fbs.

As discussed in Discord, fbs seems to demand strictly:
x.y.z

So we could go with 0.99.0 for now and increase the rightmost number for each testbuild.
Some goes for firmware.

Any thoughts on this, which parts of the software are affected by this move?

t0mpr1c3:
The rationale should be recorded in the CHANGELOG so that it's consistent going forward. And the new firmware flash dialog/function needs to be aware of it when loading firmware versions from the repo.

dl1com:
@t0mpr1c3 could you please point me to the requirements for version formatting of fbs? I did not find information about this in https://build-system.fman.io/manual/
I just want to double check the options we have for pre-release version naming (i.e. 1.0.0-rc1)

t0mpr1c3:
https://github.com/mherrmann/fbs/blob/702761412a2c8617e68ea8c398f7df725ef71a79/fbs/builtin_commands/_util.py#L72C20-L72C20

dl1com:
> 

meh, I see... ` \d+\.\d+\.\d+$` ...prepending the version number is pretty unusual and against semver.org.
I'd love to be able to use tags like 1.0.0-rc1 or similar to indicate a release candidate/alpha of an upcoming version. fbs prohibits this.

I see two possible solutions:
1. patch fbs so it accepts also other, semver compatible version number formats, or
2. use a semver-compatible format for our tags and versions like 1.0.0-rc1, also show it in our application in this format, but strip it down to 1.0.0 when writing it into the base.json for fbs, so it does not complain when running freeze. Updating an application from 1.0.0-rc1 to 1.0.0-rc2 automatically through Windows installer might not work as expected because then, they both are frozen as 1.0.0 by fbs, but for me that's no drawback as these are release *candidates*, no actual releases. For real releases, we have a difference in the version number (1.0.0 -> 1.0.1) and this should work as expected.

Therefore, I'd favor option 2.

X-sam:
My vote is also on option 2. FBS being janky isn't a good reason not to use the most common `maj.min.patch-sub` format.

X-sam:
(though that shouldn't stop us from opening an issue on FBS about this topic!)

[...]

PR #551, opened 2023-07-21
Semantic versioning

dl1com:
Desktop-software related:
- [x] Strip suffix from semver tag before writing version to fbs' base.json as fbs does not support suffixes (see discussion https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/534#issuecomment-1622452529)
- [x] Add information to CHANGELOG about the switch to semver 2.0
- [x] Check if update notification (shown on startup) needs to be changed (see https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/551#issuecomment-1646272804)

Firmware-related:
- [x] Update Github workflow in Desktop repo to download the correct firmware binary according to the manifest file with an arbitrary release name in it (currently, it requires 'v' or 'test' as prefix) - solved with #552 
- [x] Check if firmware dialogue has to be altered to show firmware version correctly (esp. in case it contains a suffix)

### DIFF
--- a/.github/workflows/build-multi-os.yml
+++ b/.github/workflows/build-multi-os.yml
@@ -61,7 +61,9 @@ jobs:
     - name: Set PACKAGE_VERSION
       run: |
         echo ${{ steps.vars.outputs.tag }} > src/main/resources/base/ayab/package_version
-        sed -i 's/PACKAGE_VERSION/${{steps.vars.outputs.tag}}/' src/build/settings/base.json
+        # remove suffix from semver tag, as fbs does not support them
+        version_without_suffix=$(echo ${{ steps.vars.outputs.tag }} | sed -E 's/^(v?[0-9]+\.[0-9]+\.[0-9]+).*$/\1/')
+        sed -i 's/PACKAGE_VERSION/$version_without_suffix/' src/build/settings/base.json
         cat src/main/resources/base/ayab/package_version
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build-multi-os.yml
```diff
@@ -61,7 +61,9 @@ jobs:
     - name: Set PACKAGE_VERSION
       run: |
         echo ${{ steps.vars.outputs.tag }} > src/main/resources/base/ayab/package_version
-        sed -i 's/PACKAGE_VERSION/${{steps.vars.outputs.tag}}/' src/build/settings/base.json
+        # remove suffix from semver tag, as fbs does not support them
+        version_without_suffix=$(echo ${{ steps.vars.outputs.tag }} | sed -E 's/^(v?[0-9]+\.[0-9]+\.[0-9]+).*$/\1/')
```
dl1com, line 65: @t0mpr1c3 the "Set PACKAGE_VERSION" step occurs in each OS job (4 times) - do you see a chance to have this in the caching job only  so we get rid of this redundancy? Same applies for "Set variables" step in each job

dl1com:
No need to update the software update notification, it seems there is nothing which prohibits the switch to semver (https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/1.0.0-dev/src/main/python/main.py#L42) it just checks if the names of the current version and of the latest release on gh are different, no parsing of version numbers.

dl1com:
Currently, the name of the firmware binary shown in the firmware flashing dialogue is hard-coded in firmware.json.
How shall we proceed here? Two possible ways:
1. Inject the firmware version name into firmware.json during the ayab-desktop release process (raises the question how to handle this in case of running from source), or
2. parsing the firmware version from the firmware file name(s) (we would have to name it appropriately, of course, i.e. ayab_uno_1.0.0-rc1.hex)

if we don't automate this,

.github/workflows/build-multi-os.yml
t0mpr1c3: We could, yes.

t0mpr1c3: I'll do it in a separate PR because it will clash with #554

t0mpr1c3: I think the `sed` code needs to be

     sed -E 's/^v?([0-9]+\.[0-9]+\.[0-9]+).*$/\1/'

(we should get rid of the initial `v`)

t0mpr1c3: But it'll be fixed in #554 anyway, so I'll resolve this now.

t0mpr1c3:
Probably (3) is easiest?

t0mpr1c3 (approved):

Issue #476, opened 2023-05-14
upgrade fbs version >=1.0.1


t0mpr1c3:
Once we have moved to semantic versioning, we will be able to upgrade `fbs` and will no longer have to patch it in the Windows build workflow.

`fbs` version 1.0.1 contains the fix to the Windows build that is currently applied using the [patch](https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/224593e94eac26feeeacb5cf53c0378035988ab1/.github/workflows/build-multi-os.yml#LL258C8-L258C114).

PR #554, opened 2023-07-24
use most recent version of `fbs`

t0mpr1c3:
Possible now that we have semantic versioning.

Closing #476


.github/workflows/build-multi-os.yml
```diff
@@ -269,16 +268,6 @@ jobs:
...
-    - name: Patch fbs package
-      shell: pwsh
-      run: |
-        $cwd = $(Get-Location).Path
-        $path = '${{ steps.py.outputs.python-path }}' -Replace '[^\\]*$', ''
-        echo $path
-        $fbs = $(Get-ChildItem -Path $path -Depth 3 -Filter 'fbs')[0].FullName + '\..\'
-        echo $fbs
-        Set-Location $fbs
-        patch -u -i $($cwd + '\windows-build\patch-fbs.diff') 'fbs/_defaults/src/installer/windows/Installer.nsi'
```
dl1com, line 281: As we don't need the file `/windows-build/patch-fbs.diff`, please remove it

requirements.txt
```diff
@@ -1,4 +1,4 @@
-fbs==0.8.6
+fbs
```
dl1com, line 1: I think it is still worth to use version pinning here, at least for maj.minor version, so `fbs==1.0`

.github/workflows/build-multi-os.yml
```diff
@@ -16,8 +16,7 @@ name: Build
 on:
   push:
     tags:
-    - 'v*'
-    - 'test*'
+    - '[0-9]+.[0-9]+.[0-9]+'
```
dl1com, line 19: This expression does not match tags with a suffix. Consider adding `(-[\S]*)?$` or similar?

requirements.txt
t0mpr1c3: Yes

.github/workflows/build-multi-os.yml
t0mpr1c3: Yes

.github/workflows/build-multi-os.yml
t0mpr1c3: fbs doesn't allow that so we'll have to truncate the tag somehow

dl1com: sed command looks good to me. if you want, you could add a short comment to explain that we truncate because of fbs. apart from that, I think we're good to go :)

requirements.txt
dl1com: Sorry, I just became aware that for allowing updates on patch level, we have to use the tilde operator (`fbs~=1.0`). With using the `==1.0` operator, it will download 1.0.0, no newer patchlevel like 1.0.1.

t0mpr1c3: Yes

.github/workflows/build-multi-os.yml
t0mpr1c3: Yes, I will add a comment when the code is changed to cache `base.json`.

requirements.txt
t0mpr1c3: Changed to `fbs~=1.2`

t0mpr1c3:
After this is merged we can delete https://github.com/AllYarnsAreBeautiful/fbs

dl1com (approved):

PR #559, opened 2023-08-01
Repair requirements.txt

dl1com:
Reinsert PyQt5 and PyInstaller requirements, although already covered by fbs 1.2.1

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ fbs @ git+https://github.com/AllYarnsAreBeautiful/fbs@1.2.1-py38
 PyQt5==5.12.3
 PyInstaller==4.1
 Pillow==7.1.2
-pyserial~=3.4
+pyserial==3.4
 sliplib==0.4.0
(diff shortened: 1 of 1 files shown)


t0mpr1c3:
> Tests run fine now.
> 
> Although these lines makes me a bit nervous:
> 
> ```
> Collecting PyQt5~=5.12 (from -r requirements.txt (line 2))
>   Downloading PyQt5-5.15.9-cp37-abi3-manylinux_2_17_x86_64.whl (8.4 MB)
> 
> Collecting PyInstaller~=4.1 (from -r requirements.txt (line 3))
>   Downloading pyinstaller-4.10-py3-none-manylinux2014_x86_64.whl (1.5 MB)
> ```
> 
> So it seems that ~= does not have the intended effect to allow only patch-level updates here. Perhaps we should stick with exact version pinning (==)?

Probably safer. Best to do it in the fbs fork as well.

t0mpr1c3 (approved):

PR #574, opened 2023-09-10
Update workflows

t0mpr1c3:
Not tested yet.


.github/workflows/build-multi-os.yml
```diff
@@ -71,7 +68,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt-get install -y avrdude libasound2-dev
-    - name: Restore pip cache
+    - name: Cache pip
```
X-sam, line 71: instead of this we can just add `cache: 'pip'` to the `setup-python` action's `with` list. [relevant doc](https://github.com/actions/setup-python#caching-packages-dependencies)

X-sam:
while we're here, we should cache/restore `/Library/Caches/Homebrew` on the mac builder, the brew stuff is the slowest part right now

t0mpr1c3:
> while we're here, we should cache/restore `/Library/Caches/Homebrew` on the mac builder, the brew stuff is the slowest part right now

not100% how to do that but I gave it the college try

PR #585, opened 2023-10-17
remove pass column of knit progress in Single Bed mode

### LINKED ISSUE #548: The "Pass" column is redundant in single bed mode
AYAB software version: test230626
Computer/OS: Intel Mac/Big Sur
Knitting machine: KH910
AYAB hardware: Interface

In the new display showing the pattern row as you knit, when you're in Single bed mode the second column always says Pass 1. How about removing it?

1. Open any image
2. Leave all the settings at default, so you're using single bed and 2 color
3. Doesn't matter whether you're set to Simulation or actually connected.
4. Click Knit and let it run a few rows.

Columns 1 and 3 contain useful information, but column 2 always says "Pass 1" which isn't very informative.

Screen Shot 2023-07-01 at 11 54 40 PM

t0mpr1c3:
Closes #548

### DIFF
--- a/src/main/python/ayab/knitprogress.py
+++ b/src/main/python/ayab/knitprogress.py
@@ -18,7 +18,7 @@
 #    https://github.com/AllYarnsAreBeautiful/ayab-desktop
 
 from PyQt5.QtCore import Qt, QCoreApplication, QRect, QSize
-from PyQt5.QtWidgets import QTableWidget, QTableWidgetItem, QLabel, QSizePolicy, QAbstractItemView, QWidget, QHBoxLayout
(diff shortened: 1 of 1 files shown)


dl1com (approved):

dl1com:
@t0mpr1c3 it seems that there is still a run-pytest action for  Python 3.6 in the test pipeline. It never succeeds (obviously). Could you please check and remove it?

t0mpr1c3:
> it seems that there is still a run-pytest action for Python 3.6 in the test pipeline. It never succeeds (obviously). Could you please check and remove it?

I can't find it! ):

PR #619, opened 2024-02-24
bump PyQt5 to version 5.15.10

t0mpr1c3:
Not immediately necessary, but the upgrade seems to work on Windows.


dl1com (approved):

dl1com:
If it seems to work, no reason to wait...let's see if it breaks :D

Issue #614, opened 2024-02-20
[BUG] When Knit Side Image is checked, the image on the knit side should be the same as the saved image.


Adrienne200:
**Environment**
AYAB software version: 0.99 RC2
Computer/OS: Windows 7
Knitting machine: KH910
AYAB hardware: Shield or interface

Companion to #613, when Knit Side Image is checked, the image should be displayed backwards in both screen images, and knit backwards as you see it as you knit, so that it is the right way round on the knit side.

1. AYAB 1.0 RC2 (with the firmware reflashed and the correct machine set in the Prefs)
2. Open the test image uC3, and bring needles L15 - R15 into work.
3. Leave all the settings at default, except do check Knit Side Image.
4. Start AYAB as usual, waiting for the beep after crossing the left turn mark.
5. Observe which way round the image appears on the actual knitting, which side does the tail of the mu character appear?

Expected behavior:
The UI as shown in the screenshot is partially correct for the Knit Side Image box checked. Both images are flipped, that is correct, but two changes:
 - the tail of the mu character should appear on the right, the green side, to match the images as displayed, but it currently appears on the left.
 - the orange and green bars should not be reversed, leave them at their usual setting.

![Mirroring 2](https://github.com/AllYarnsAreBeautiful/ayab-desktop/assets/24358225/50b237c4-248c-4469-9d60-bcc959a97fb3)

Adrienne200:
The UI is good now, (0.99.0-rc4, which installs fw rc3) but the actual knitting is still wrong. For this case the tail of the mu character should appear on the green/right needles, which matches the UI, and so that the text will be the right way round on the knit side.

Adrienne200:
Same at rc6. Companion to #613, the UI is good, just needs the actual knitting flipped to match. All settings at default except Knit Side Image is checked. The image flips correctly on the screen -  the actual knitting should flip too.

jonathanperret:
I don't have hardware yet to confirm in a real setting, but I added logging to monitor the bytes written to the line data in simulation. Running the current state of `1.0.0-dev` (5037932d909a7af5a5cdcba37736a0665f2cf76a) here are the `cnfLine` packets written out when knitting the `µC3` pattern with all default settings:

```
nr=0 color=0 flags=0x0	line_data=00000000000000000000000000000000000000000000000000000000000000000000000000000000000001100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
[...]

Issue #613, opened 2024-02-19
[BUG] Default view should knit the image as you see it on the screen


Adrienne200:
AYAB software version: 0.99 RC2
Computer/OS: Windows 7
Knitting machine: KH910
AYAB hardware: Shield or interface

Following from the conversation in #604, I'm looking at the simplest case, all default settings, open an asymmetrical image, what should it look like? On the screen, and which needles should knit? I think this should be the WYSIWYG case, open the image so it looks exactly as you saved it, and knit it out so that the image as you see it appears on the side facing you as you knit, the purl side.

1. AYAB 1.0 RC2 (with the firmware reflashed and the correct machine set in the Prefs)
2. Open the test image uC3, and bring needles L15 - R15 into work.
3. Leave all the settings at default, (Knit Side Image is unchecked)
4. Start AYAB as usual, waiting for the beep after crossing the left turn mark.
5. Observe which way round the image appears on the actual knitting, which side does the tail of the mu character appear?

Expected behavior:
The UI as shown in the screenshot is, I believe, correct as it stands for the non-KSI case, but I want it to finish the job by having the actual knitting also match the image as you look at it, so the tail of the mu character should appear on the left, the yellow side, but it currently appears on the right.

![unmirroring](https://github.com/AllYarnsAreBeautiful/ayab-desktop/assets/24358225/2b8527a5-50ee-4003-9917-9173ff68c287)

t0mpr1c3:
Thanks for this.

One supplementary question: what do we want "align left" and "align right" to mean?

In the current version of the software (Jan 2, commit 8bdb8a6), with the "knit side" box unchecked, looking at the purl side:
- When alignment is set to "left", that aligns on the LHS of the screen image. But on the knitted article, looking at the knit side, this will result in an image that is aligned to the RHS of the garment.
- With alignment set to "left", when we check the "knit side image" checkbox, the image is now aligned to the RHS of the screen.

I'm not sure what the original Brother software does, but the current behaviour seems wrong. If "align left" means "align on the LHS of the screen", then that should happen whether or not the "knit side image" checkbox is checked. If "align left" means "align in the LHS of the finished garment", then that is the reverse of what is happening at the moment.

Adrienne200:
The behavior is still the same in 0.99.0-rc4 (Mac)

(The Alignment feature is separate. This is about the basic behavior with the default center alignment.)

citizenkate:
[...]

PR #621, opened 2024-02-25
WYSIWYG #613/#614

t0mpr1c3:
Fixes the concerns raised in #613 and #614, and I think things should now work for centre alignment, but I still have concerns that left/right alignment are not quite right.


dl1com:
@t0mpr1c3 Could you please fix the branch conflicts? I am not confident enough to fix them on the fly in the web editor...

t0mpr1c3:
@dl1com should be good to go

dl1com (approved):

Issue #635, opened 2024-03-13
[BUG] Appimage 0.99.0-rc4 does not launch


dl1com:
When trying to run AppImage 0.99.0-rc4 in Debian testing, it throws an error regarding Python 3.8. (Although this release is/should be based on 3.11)

```bash
chmod a+x AYAB-0.99.0-rc4-x86_64.AppImage 
./AYAB-0.99.0-rc4-x86_64.AppImage 
/tmp/.mount_AYAB-0NMKAkp/AppRun: line 14: opt/python3.8/bin/fbs: No such file or directory

```

dl1com:
Working in 1.0.0-beta1

PR #636, opened 2024-03-13
Fix AppImage Python Version

X-sam:
idk much about desktop linux appimages but this theoretically addresses #635


dl1com:
Lgtm, worth a try

dl1com (approved):

PR #637, opened 2024-03-14
Add typing

X-sam:
This adds fairly comprehensive typing to ayab-desktop. 
In a local environment, you can test by running `mypy src/main/python` .

PR #638, opened 2024-03-16
Be More Opinionated About Linting

X-sam:
This switches us from `yapf` to `black` (for deterministic formatting on github actions runners), formats the ever-loving shizz out of the codebase, and resolves all issues flake8 raises. It also updates our pre-commit config to match the tools we're using- `flake8`, `black`, and `mypy`.
This is dependent on #637

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+max-line-length = 80
+extend-select = B950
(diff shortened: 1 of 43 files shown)


dl1com (approved):
Lots of changes 😅 lgtm

PR #654, opened 2024-04-07
Reset number of colors and start row on loading new image

### LINKED ISSUE #599: [BUG] After loading a new image, the number of colours and start row do not reset to their initial values
**Environment**
AYAB software version: current
Computer/OS: all
Knitting machine: -
AYAB hardware: -

**Describe the bug**
After you load a new image, the options panel is supposed to reset to the default settings. (It doesn't: that's a bug, and I have submitted #598 to fix it.) But there are other settings that don't have defaults, such as number of colours, and start row. These should also be reset (to the starting values, 2 and 1 respectively) after loading a new image because the new image might have fewer colours/rows than the previous image, so the values could be invalid.

**To Reproduce**
1. Load a three-colour image.
2. Change number of colours to 3.
3. Load a two-colour image.

**Expected behavior**
The number of colours should reset to 2. The start row should reset to 1.

t0mpr1c3:
Closes #599

### DIFF
--- a/src/main/python/ayab/ayab.py
+++ b/src/main/python/ayab/ayab.py
@@ -190,6 +190,10 @@ class GuiMain(QMainWindow):
         """Flip image horizontally."""
         self.scene.reverse()
 
+    def got_image(self) -> None:
+        """Reset image."""
+        self.engine.config.reset_image()
+
     def update_start_row(self, start_row: int) -> None:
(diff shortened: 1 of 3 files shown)


dl1com:
Setting to draft as it seems there is still discussion ongoing in #599, plus a rebase of the branch is necessary before merge.

PR #669, opened 2024-06-28
Make knitting match what you see

### LINKED ISSUE #613: [BUG] Default view should knit the image as you see it on the screen

### LINKED ISSUE #614: [BUG] When Knit Side Image is checked, the image on the knit side should be the same as the saved image.

jonathanperret:
The most important change happens in `engine.py`: we now only flip the image vertically, instead of rotating it 180 degrees when preparing the pattern.

In `KnitProgress` we remove the left-to-right reversal that was done to compensate for the unnecessary mirroring.

In addition, while researching `bitarray` I discovered that we were using a discontinued fork of the official `bitarray` package. As far as I can tell there are no undesirable effects of switching back to the `bitarray` package, and since it now guarantees that arrays are initialized with zeroes, a couple uses can be simplified.

Fixes #613. Also fixes #614 I think, since the "Knit side image" checkbox appears to work, only the output needed to be unflipped.

Edit: please disregard the AI attempt at summarizing the changes below, it is wrong on multiple counts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the initialization of bitarrays to ensure compatibility and performance improvements.

- **Documentation**
  - Added detailed documentation to the `cnf_line_API6` function to clarify bit control for the leftmost needle.

- **Enhancements**
  - Improved image handling in the `knit_config` method by changing image transformation to a top-to-bottom flip.
  - Optimized bitarray initialization in multiple modules for better efficiency.

### DIFF
--- a/requirements.build.txt
+++ b/requirements.build.txt
@@ -4,7 +4,7 @@ PyInstaller==6.4.0
 Pillow==10
 pyserial==3.5
 sliplib==0.6.2
-bitarray-hardbyte==2.3.8
+bitarray==2.9.2
 simpleaudio==1.0.4
 wave==0.0.2
 numpy==1.26.4
--- a/src/main/python/main/ayab/engine/communication.py
+++ b/src/main/python/main/ayab/engine/communication.py
@@ -174,6 +174,8 @@ class Communication(object):
           color (int): The yarn color to be sent.
           flags (int): The flags sent to the controller.
           line_data (bytes): The bytearray to be sent to needles.
+              The leftmost needle is controlled by the least-significant bit of
+              the first byte.
         """
         if self.__ser is None:
             return
--- a/src/main/python/main/ayab/engine/control.py
+++ b/src/main/python/main/ayab/engine/control.py
@@ -265,7 +265,7 @@ class Control(SignalSender):
     def select_needles_API6(
         self, color: int, row_index: int, blank_line: bool
     ) -> bitarray:
-        bits = bitarray([False] * self.machine.width, endian="little")
+        bits = bitarray(self.machine.width, endian="little")
 
         # select needles flanking the pattern
         # if necessary to knit the background color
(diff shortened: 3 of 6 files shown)


dl1com (approved):

jonathanperret:
Thanks @dl1com !

PR #672, opened 2024-06-28
Fix refreshing alignment in scene view

### LINKED ISSUE #671: The alignment selector has no apparent effect
To reproduce (tested on `0.99.0-rc6` on an M1 Mac):
- open AYAB
- load an image less than 200 pixels wide
- increase start and stop needle values to give the pattern space to be moved left or right
- change the Alignment value from Center to either Left or Right

Expected behavior: the main view reflects the pattern moving to the left or right of the selected area.

Observed behavior: nothing visible happens.

https://github.com/AllYarnsAreBeautiful/ayab-desktop/assets/300823/4d2deea3-790b-4c4b-982f-04d1a04f3789

jonathanperret:
Because of a typo, the `alignment_updater` was never called.

Fixes #671.

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected the alignment update feature to ensure the proper parameter is emitted during the signal sending process.

### DIFF
--- a/src/main/python/main/ayab/signal_sender.py
+++ b/src/main/python/main/ayab/signal_sender.py
@@ -81,7 +81,7 @@ class SignalSender(object):
         self.__signal_receiver.needles_updater.emit(start_needle, stop_needle)
 
     def emit_alignment_updater(self, alignment: Alignment) -> None:
-        self.__signal_receiver.alignment_updater.emit(Alignment)
+        self.__signal_receiver.alignment_updater.emit(alignment)
 
     def emit_image_resizer(self) -> None:
         self.__signal_receiver.image_resizer.emit()


X-sam:
Why did the original code make it past the type checker, I wonder

jonathanperret:
> Why did the original code make it past the type checker, I wonder

The `emit` method on `SignalInstance` has untyped arguments:

```
class SignalInstance(object):
    def connect(self, slot: object, type: PySide6.QtCore.Qt.ConnectionType = ...) -> PySide6.QtCore.QMetaObject.Connection: ...
    def disconnect(self, slot: Optional[object] = ...) -> None: ...
    def emit(self, *args: Any) -> None: ...
```

dl1com (approved):

PR #673, opened 2024-07-07
Fix macOS build with new bitarray package

jonathanperret:
The `bitarray` package which was updated in #669 now has single-architecture macOS binary wheels available, which causes the app packaging to fail complaining that its native library is not a universal binary (see https://github.com/jonathanperret/ayab-desktop/actions/runs/9823270540/job/27121238943).

This PR applies a similar workaround as was employed to fix a similar problem for `numpy` and `Pillow`. In the `bitarray` case however, we don't have to merge two single-architecture wheels since fortunately a universal wheel is available, it's just not picked up by default by `pip install`.

I also improved the code that downloads the `numpy` and `Pillow` wheels a bit, so that it will track the version of these packages mentioned in `requirements.build.txt` instead of a hardcoded version.

## Summary by CodeRabbit

- **Chores**
	- Improved build workflow to fetch platform-specific binaries for numpy and Pillow dependencies.
	- Added platform-specific download and installation steps for the bitarray dependency.

### DIFF
--- a/.github/workflows/build-multi-os.yml
+++ b/.github/workflows/build-multi-os.yml
@@ -313,15 +313,23 @@ jobs:
         run: |
           brew install avrdude
           brew install qt
-          python -m pip install delocate
-          python -m pip download --only-binary=:all: --platform macosx_10_09_x86_64 numpy==1.26.4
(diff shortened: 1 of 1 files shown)


dl1com (approved):

dl1com:
Thanks 🥳

PR #677, opened 2024-07-12
Remove requirement for separate Qt installation

jonathanperret:
The only reason to require a standalone Qt installation was the `lrelease` tool used to convert the translation files.

It turns out PySide6, which we install through `pip`, already includes this tool under the name `pyside6-lrelease`.

This PR changes the call from `lrelease` to `pyside6-lrelease`, and removes the installation of Qt from the GitHub Actions workflow and from the development setup instructions.

Except on Linux, where the Qt libraries which are bundled in PySide6 have themselves dependencies on system libraries such as `libgl1` and a dozen others. In that case it seemed simpler to keep installing a system Qt 6 package to pull these system dependencies in, even though none of the Qt components installed system-wide will be used.

## Summary by CodeRabbit

- **New Features**
  - Upgraded Python environment setup to the latest version, enhancing compatibility and performance.
  - Streamlined installation process for Qt6 system dependencies for better efficiency.

- **Documentation**
  - Improved clarity and conciseness of the installation instructions in the README, focusing on essential steps for users.

- **Chores**
  - Updated command for generating binary files to ensure compatibility with newer versions of the PySide6 framework.

### DIFF
--- a/.github/workflows/build-multi-os.yml
+++ b/.github/workflows/build-multi-os.yml
@@ -104,8 +104,9 @@ jobs:
           (steps.e-cache.outputs.cache-hit    != 'true') ||
           (steps.qm-cache.outputs.cache-hit   != 'true') }}
         run: |
-          sudo apt install -y qt6-base-dev qt6-tools-dev-tools qt6-l10n-tools
-          PATH=/usr/lib/qt6/bin:$PATH bash setup-environment.ps1
+          # Install qt6 packages to bring in system dependencies only
+          sudo apt install -y qt6-base-dev qt6-tools-dev-tools
+          bash setup-environment.ps1
       - name: Cache gui files (1)
         if: ${{ (steps.gui1-cache.outputs.cache-hit != 'true') }}
         uses: actions/cache/save@v3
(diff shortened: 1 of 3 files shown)


jonathanperret:
Since the Build workflow doesn’t trigger for PRs here’s a run on my fork on the same commit, to demonstrate everything works: https://github.com/jonathanperret/ayab-desktop/actions/runs/9908222560

X-sam (changes requested):
Can you break the changes out into multiple PRs? I'm ok removing QT, but I am less prepared for testing the updates to actions at the moment, and I added the type generation for a reason so I would rather we don't remove it without further discussions and testing.

jonathanperret:
> Can you break the changes out into multiple PRs? I'm ok removing QT, but I am less prepared for testing the updates to actions at the moment, and I added the type generation for a reason so I would rather we don't remove it without further discussions and testing.

You're right, I got a bit carried away there. I've pared down the PR to just the removal of Qt. As I added in the PR description, on Linux we need to somehow get Qt's system dependencies installed, so I left `apt install -y qt6-base-dev qt6-tools-dev-tools` in to get these.

dl1com:
@X-sam as @jonathanperret did some changes to the PR, do you think we can merge it now?

X-sam (approved):

PR #679, opened 2024-07-23
Improve serial read latency and throughput

jonathanperret:
## Issue

The current implementation of `Communication.read_API6` always blocks until an unrealistic 1000 bytes have been read from the serial port, causing the `__ser.read()` call to end with the configured timeout (100ms) for every message. This imposes a latency of up to 100ms on any given message coming from the firmware, since even after receiving a complete SLIP packet, the `read()` call will block waiting to reach 1000 bytes.

Also, in the case where we have received multiple messages in a single read, after being extracted by the SLIP decoder they are all be enqueued in `rx_msg_list` but only the first one is returned from `read_API6`. To access the next message, the calling thread has to wait again for the 100ms timeout to expire. This limits message throughput to 10 messages per second.

This is likely to be the cause of the delay reported in https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/257#issuecomment-441147394 . In continuous reporting mode, the firmware can probably generate more than 10 messages per second; a 115200 bps link should support roughly a thousand 10-byte messages per second, but it is accidentally hobbled by the problem described above.

## Proposed solution

By waiting for a single byte to arrive, we make sure to unblock the `read()` call as soon as any data is available, while avoiding a CPU-heavy busy loop (if the firmware is completely silent, the read() call will block for 100ms each time, like it did before). To avoid extra read passes, after we get one byte we try to pull as many bytes as are available in the serial buffer. This takes care of the latency issue.

To improve throughput in the case where we have received multiple messages in a single read, we avoid blocking on a wait for more data by only going to the serial port once we have consumed all already-read messages.

## Demonstration

It is difficult to show the impact of improved latency on actual knitting since artificial pauses were introduced to compensate for the issue, so here's a visual demonstration of the impact on the hardware testing mode (note that this was recorded using my [work-in-progress branch](https://github.com/jonathanperret/ayab-desktop/tree/fix-hw-test) where hardware test mode is functional; it won't work on `1.0.0-dev` currently).

In both cases the firmware is running on an Arduino Uno R3. The serial port bitrate is set at 115200 bps as usual.

Here's the experience without the change in this PR:

https://github.com/user-attachments/assets/a07294ea-84eb-4a25-adfe-ca4ce410b55c

Now with the change proposed in this PR:

https://github.com/user-attachments/assets/c45c3baa-9282-44d4-9283-7bb2503129a4

## Summary by CodeRabbit

- **New Features**
	- Improved the efficiency of serial data communication handling.
	- Enhanced responsiveness to incoming data by refining data reading logic.
  
- **Bug Fixes**
	- Added checks to prevent attempts to read from a non-existent serial connection, reducing errors.

- **Documentation**
	- Updated the method documentation for better clarity on functionality.

### DIFF
--- a/src/main/python/main/ayab/engine/communication.py
+++ b/src/main/python/main/ayab/engine/communication.py
@@ -209,14 +209,34 @@ class Communication(object):
         return msg, Token.unknown, 0
 
     def read_API6(self) -> Optional[bytes]:
-        """Read data from serial as SLIP packet."""
-        if self.__ser:
-            data = self.__ser.read(1000)
+        """Read data from serial, return the next SLIP packet"""
+
+        if self.__ser is None:
+            return None
+
+        # If we already have messages pending from previous serial reads,
+        # do not bother waiting on the serial port. The data will be safely
+        # buffered by the OS until we have consumed the messages already
+        # received.
+        if len(self.rx_msg_list) == 0:
+            # This will block until the timeout configured in the Serial constructor
+            # if no data is available (to avoid busy-waiting), but return as soon as
+            # a single byte is available to read.
+            data = self.__ser.read(1)
+
+            # More bytes may have become available simultaneously: grab them now.
+            if self.__ser.in_waiting > 0:
+                data = data + self.__ser.read(self.__ser.in_waiting)
+
+            # Send everything we received to the SLIP decoder and enqueue any messages
+            # it extracted from the data so far.
             if len(data) > 0:
                 self.rx_msg_list.extend(self.__driver.receive(data))
-            if len(self.rx_msg_list) > 0:
-                return self.rx_msg_list.pop(0)  # FIFO
-        # else
+
+        # Now, return the oldest message we have in the queue.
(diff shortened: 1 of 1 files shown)


dl1com:
Great improvement! ...as this is a crucial part of the software, I would love to see this PR being tested on a machine before getting merged...  @AllYarnsAreBeautiful/ayab-contributors any volunteers?

jonathanperret:
> Great improvement! ...as this is a crucial part of the software, I would love to see this PR being tested on a machine before getting merged... @AllYarnsAreBeautiful/ayab-contributors any volunteers?

I 100% agree that this should be tested in a real knitting scenario, which I can't do right now unfortunately.

If anyone is up for it here's a build with the code from this PR: https://github.com/jonathanperret/ayab-desktop/releases/tag/v1.0.0-20240723-fix-latency

dl1com (approved):

dl1com:
Tested by @Adrienne200 
> I've tried out your latency-reduction build on an Intel Mac. No obvious differences, seems the same as previous builds.

Issue #680, opened 2024-07-24
[BUG] macOS: a crash report pop-up appears on exiting the app


jonathanperret:
**Environment**
AYAB software version: 1.0.0-dev
Computer/OS: macOS 14.2 on M1

**Describe the bug**

The AYAB app causes a crash report window ("AYAB quit unexpectedly") to pop up after quitting.

**To Reproduce**
1. Launch a recent build of AYAB, for example https://github.com/AllYarnsAreBeautiful/ayab-desktop/releases/tag/v1.0.0-test-knit-display
	⚠️ Please check an actual build as downloaded from GitHub or produced locally with `fbs freeze`, not running from source with `fbs run`.
	⚠️ Launch the app by double-clicking its icon in the Finder; I've noticed that running it from a terminal can suppress the crash report pop-up (in particular, launching from the Visual Studio Code terminal disables the crash report window for me).
2. Quit the app by doing one of the following (when testing that the bug is fixed, please try each of these at least once):
	- Selecting the `Quit` item from the `File` menu in the AYAB window
	- Select the `Quit` item from the `AYAB` menu in the macOS menu bar
	- Pressing the `Cmd+Q` shortcut while AYAB is the active application
	- Right-clicking or Ctrl-clicking the AYAB icon in the Dock and selecting the "Quit" menu item
	- Closing the main AYAB window by clicking the red close button in the top-left corner
3. Potentially wait a few seconds while macOS collects the crash report information.
4. See a macOS crash report window pop up:

<img width="1112" alt="AYAB-crash-report" src="https://github.com/user-attachments/assets/f66f0118-a4ed-4206-ba81-c859602f72b9">

If you don't see a crash report window, please check the following before marking the bug as fixed:
 - Make sure you don't have an existing crash report window open in the background, as macOS will not open a second crash report window for the same app while one is already open. This is made tricky by the fact that this window does not have an associated icon on the Dock. It does appear on Mission Control/Exposé though, just make sure it is not open on another desktop or screen.
 - It sems that sometimes macOS decides that you've seen enough of these crash report windows and stops showing them. Clearing the `~/Library/Logs/DiagnosticReports` folder where the crash reports are stored may help, or rebooting. When in doubt, I found it helps to keep a known-affected build of AYAB around so that you can run it and check that the crash report window still appears.

**Expected behavior**
No crash report window should pop up when exiting the app.

**Additional context**

[...]

PR #681, opened 2024-07-24
Remove audio player thread to fix crash on exit

jonathanperret:
## ⚡ Summary

This PR fixes issue #680 by removing the audio player thread which is the cause of the crash on exit, due to not being exited properly before app termination.

## ℹ️ Details

The current code spawns a dedicated thread (introduced in #382) to handle playing sound effects.

This is unnecessary as the `simpleaudio` API already is non-blocking, i.e. calling `wave_obj.play()` returns immediately while the sound plays in the background (see https://simpleaudio.readthedocs.io/en/latest/capabilities.html#asynchronous-interface). At the worst, loading the sound file, which is synchronous, could block the calling thread for a few milliseconds but that happens only once and is unlikely to be an issue.

The `AudioWorker.play` method does have a `blocking` argument that is supposed to support blocking the caller until the sound has finished playing, but this is never used in the current codebase.

But more conclusively, it turns out that due to an oversight, practically no code currently executes on the audio player thread. The `AudioWorker` object is assigned to a new thread, but because the `AudioPlayer.play` method calls `__worker.play` without going through a signal:
https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/bc96d5c3e3f955fa75551ec44c82e966e0981112/src/main/python/main/ayab/audio.py#L91-L92

…the `AudioWorker.play` method actually executes on the thread that called `AudioPlayer.play`. In practice this is the UI thread, because `audio.play` is called through a signal on `SignalReceiver`, which lives on that thread. Thus, the time-sensitive `engine` thread is already isolated from any slowness in reading or playing audio.

This commit reduces the `AudioPlayer` class to a plain Python object whose only responsibilities are to load and play the application's sound effects. It also simplifies the file loading logic by removing the direct call to the `wave` library in favor of the `from_wave_file` helper method offered by `simpleaudio`, which in practice uses the exact same call to the `wave` library.

A notable benefit of removing this thread is that the application no longer crashes with an `abort` call upon exit (issue #680), caused by improper termination of the audio thread.

## 🔬 How to test

- Knit a pattern from start to finish; observe that the _start_, _new line_ and _end_ sound effects are still played correctly. Note that (at least for me) when knitting in "Simulation" mode, the per-line sound effect is not heard; this is not a regression introduced by this PR since it already occured before and needs to be investigated separately.
- Follow the instructions in #680 to ascertain that a crash no longer happens on exit.

## Summary by CodeRabbit

- **New Features**
	- Simplified audio playback functionality for a more efficient user experience.
  
- **Bug Fixes**
	- Removed unnecessary complexity in audio handling, which may improve responsiveness.

- **Refactor**
	- Streamlined the `AudioPlayer` class to enhance performance and maintainability by removing threading logic.
	- Updated signal connections to improve the handling of audio playback. 

- **Chores**
	- Removed unused dependencies from the project to minimize bloat.

### DIFF
--- a/src/main/python/main/ayab/audio.py
+++ b/src/main/python/main/ayab/audio.py
@@ -20,29 +20,24 @@
 """Standalone audio player."""
 
 from __future__ import annotations
-import logging
 from os import path
 
 import simpleaudio as sa
-import wave
 
-from PySide6.QtCore import QObject, QThread
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from .ayab import GuiMain
 
 
-class AudioWorker(QObject):
-    def __init__(self, parent: GuiMain):
-        super().__init__()
-        self.__dir = parent.app_context.get_resource("assets")
-        self.__prefs = parent.prefs
+class AudioPlayer:
+    def __init__(self, gui: GuiMain):
+        self.__dir = gui.app_context.get_resource("assets")
+        self.__prefs = gui.prefs
         self.__cache: dict[str, sa.WaveObject] = {}
 
-    def play(self, sound: str, blocking: bool = False) -> None:
-        """Play audio and wait until finished."""
-        # thread remains open in quiet mode but sound does not play
+    def play(self, sound: str) -> None:
+        """Play audio."""
         if self.__prefs.value("quiet_mode"):
             return
         # else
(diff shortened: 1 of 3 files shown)


src/main/python/main/ayab/audio.py
```diff
@@ -20,73 +20,43 @@
...
             return
         # else
-        play_obj = wave_obj.play()
-        if blocking:
-            # wait until sound has finished before returning
-            play_obj.wait_done()
+        wave_obj.play()
 
-    def __wave(self, sound: str) -> Optional[sa.WaveObject]:
+    def __wave(self, sound: str) -> sa.WaveObject | None:
```
X-sam, line 50: Why this change?

src/main/python/main/ayab/audio.py
```diff
@@ -20,73 +20,43 @@
...
         if sound not in self.__cache:
-            self.__cache[sound] = self.__load_wave(sound)
+            wave_object = self.__load_wave(sound)
+            if wave_object is None:
+                return None
+            self.__cache[sound] = wave_object
         return self.__cache[sound]
 
-    def __load_wave(self, sound: str) -> Optional[sa.WaveObject]:
+    def __load_wave(self, sound: str) -> sa.WaveObject | None:
```
X-sam, line 59: Same q as above

src/main/python/main/ayab/audio.py
jonathanperret: It is my understanding that `T | None` is now the recommended syntax over the older `Optional[T]`, e.g. see https://stackoverflow.com/a/69440627 .

I think I changed this because my IDE suggested it. I can revert this change (and the one below) if you don't like it, or if you prefer to upgrade all occurrences of `Optional` in a separate PR.

src/main/python/main/ayab/audio.py
jonathanperret: See my reply above 😉

X-sam (approved):

PR #693, opened 2024-07-30
Fix hardware test dialog

jonathanperret:
This PR fixes the hardware test dialog, in simulation as well as with an actual Arduino connected. Note that this requires a recent firmware build as stack overflow issues prevented the firmware from reliably entering test mode without crashing.

## Summary by CodeRabbit

- **New Features**
	- Introduced new command options including `setSingleCmd`, `setAllCmd`, `readEOLsensorsCmd`, `readEncodersCmd`, `autoReadCmd`, and `autoTestCmd`.
	- Expanded the command list in the Hardware Test Dialog to improve user interaction with new functionalities.

- **Bug Fixes**
	- Enhanced error handling in the `write_API6` method to prevent potential issues when the serial connection is invalid.

- **Improvements**
	- Streamlined command handling methods for better clarity and functionality.
	- Updated button interactions to support new commands and improve usability.


src/main/python/main/ayab/hw_test.py
```diff
@@ -43,7 +43,19 @@
 class HardwareTestDialog(QDialog):
     """Console for hardware tests."""
 
-    commands = ["help", "send", "beep", "read", "auto", "test", "quit"]
+    commands = [
+        "help",
+        "send",
+        "beep",
+        # "setSingle",     # called when checking solenoids
+        # "setAll",        # no UI to handle argument
```
X-sam, line 51: why are these commented out

jonathanperret: I initially copied them all from the `Token` definition. Most test commands in the API get a button in the UI, except for these few which either can't really be called by the current UI (`setAll`) or are called using an alternative UI which isn't a simple button (`setSingle`). The intent of my comments was to explain why they were commented out, but I admit it's not very clear.

What do you think, should I just delete those lines?

X-sam: yeah I'd delete them

X-sam:
Some comments-
Beep doesn't seem to work (I assume that's a problem for `ayab-firmware` though)
We need to add a `stop` button to the dialog- while `quit` does issue `stop`, it's kinda annoying to have to reopen the window if you want to switch test commands.
`Quit` should probably be `Close`, as it's a modal not its own app.

jonathanperret:
> Looks good and works well, thanks for getting this sorted! Some comments- Beep doesn't seem to work (I assume that's a problem for `ayab-firmware` though)
It is, I’ll look into it over there.

> We need to add a `stop` button to the dialog- while `quit` does issue `stop`, it's kinda annoying to have to reopen the window if you want to switch test commands.

If you’re talking about the “autoXxx” buttons, they’re supposed to stay highlighted while that test is running, and you can stop the test (i.e. emit `stopCmd`) by re-clicking the same button to toggle it. Isn’t that happening for you?

> `Quit` should probably be `Close`, as it's a modal not its own app.
Good catch, I’ll fix this.

X-sam:
It is not clear in the UI that they are toggles

src/main/python/main/ayab/hw_test.py
jonathanperret: done.

jonathanperret:
> > `Quit` should probably be `Close`, as it's a modal not its own app.
> 
> Good catch, I’ll fix this.

On second thought… I didn't have the dialog in front of me when I replied, sorry. `quit` here is one of the test command buttons, the command is called `quit` in the API so it would be inconsistent to call it `close` IMHO. It's not even in the proper place for a dialog "Close" button. I'll concede that the fact it also closes the window is somewhat confusing. Though the window can always be closed using the regular close button in the title bar.

Likewise with the `autoTest`/`autoRead` buttons, I agree it's not very discoverable that they are toggles.

The thing is, I didn't design any of this UI. I just tried to make it work while changing as little of it as possible, because I'm afraid tweaking the UI is not going to help us get a release out soon.

How about using this as it is now for the next release, it can't be worse than the current broken test mode that simply hangs?

X-sam:
Fine

X-sam (approved):
This working is super helpful for debugging, turns out!

jonathanperret:
> - Beep doesn't seem to work (I assume that's a problem for `ayab-firmware` though)

@X-sam the fix for this is here, enjoy: https://github.com/AllYarnsAreBeautiful/ayab-firmware/pull/199

Issue #698, opened 2024-08-19
[FEAT] Keep the screen awake while knitting


mathPi:
Issue I had:
- While knitting a big piece, my computer automatically passed in screen save mode and thus my pattern was not updating anymore on the knitting machine

Proposed solution:
I could just change the screen saver parameters to avoid that to append again but I think the best solution would be to lock the screen saver while knitting (like when watching a movie for instance).

This python package seems to do the job: [wakepy](https://pypi.org/project/wakepy/)

I just don´t know where to put it in Ayab code. If someone can help me, I can propose a PR.

mathPi:
I tried a PR #699, what do you think about it ?

Adrienne200:
I asked about this in #258 , at the time it was decided not to do it. I still want this feature, so thanks for bringing it up again. As we have the Prefs panel, it may make more sense now.

mathPi:
Ohh great, look at #699, I added the opiton in the preference screen.

jonathanperret:
As I mentioned over in #699 I'm -1 on this being a preference — let's have it always enabled. Also, after talking to @dl1com on Discord, there seems to have been a misunderstanding over #258 — I think he's OK with preventing the computer from going to sleep, just not with trying to resume knitting after the computer _has_ gone to sleep and back (@dl1com  feel free to correct me if I've misrepresented your position).

jonathanperret:
Fixed by #699.

PR #699, opened 2024-08-19
feat: keep the computer awake while knitting #698

mathPi:
## Summary by CodeRabbit

- **New Features**
  - The application now keeps your screen active during extended operations, ensuring uninterrupted performance.
  - Process notifications have been refined to provide clearer updates on completion and cancellation statuses.
  - A new dependency, `wakepy`, has been integrated to enhance screen management during operations.

### DIFF
--- a/src/main/python/main/ayab/engine/engine.py
+++ b/src/main/python/main/ayab/engine/engine.py
@@ -24,6 +24,7 @@ from PIL import Image
 
 from PySide6.QtCore import QCoreApplication, Signal
 from PySide6.QtWidgets import QDockWidget
+from wakepy import keep
 
 from .. import utils
 from ..machine import Machine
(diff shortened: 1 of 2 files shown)


X-sam (changes requested):
We should put this behind a setting, so users can turn it off.

mathPi:
> We should put this behind a setting, so users can turn it off.

Ok, I´ll try to add it. In the preference menu ?

mathPi:
I added the option in the preferences. It tested manually, it is working. I think I´ll need to add a translation for the message ?

jonathanperret:
> We should put this behind a setting, so users can turn it off.

Are we sure about this needing to be a setting? Who actually wants their computer to go to sleep while knitting?

Personally I'd rather not have the UI cluttered with a toggle no-one is reasonably going to use. Also, it makes the code uglier because the context manager can't be used anymore.

On another note, we just froze features for `1.0.0` so @mathPi your contribution may have to wait until the real release is out before it is integrated. Hopefully that won't be too long 🤞

src/main/python/main/ayab/engine/engine.py
```diff
@@ -166,37 +168,51 @@ def run(self, operation: Operation) -> None:
...
-
-        self.control.stop()
+        keep_mode = None
+        if self.preferences.value("keep_awake"):
+            self.__logger.debug("Keeping screen awake activated")
+            # Activate the keep mode only if the preference is activated
+            keep_mode = keep.presenting(on_fail="pass")
+            keep_mode.__enter__()
+
+        try:
```
jonathanperret, line 178: I still think it shouldn't be a preference, but if it has to be one, at least it can be made a bit more palatable with `nullcontext` that is designed for such cases:

```suggestion
        from contextlib import nullcontext # at the top
        …
        with keep.presenting(on_fail="pass") if self.preferences.value("keep_awake") else nullcontext():
```

Then you can get rid of the `try…finally` block.

mathPi: If it is okay for others, I'd prefer removing the settings ...
But if not yes that's a nicer solution.

dl1com:
@mathPi  as 1.0 has been released, we now could proceed with merging your PR.
Can you have a look at the code conflicts which arised in the meantime and fix them?
Thanks a lot!

src/main/python/main/ayab/engine/engine.py
dl1com: What was the conclusion of this discussion? Having it as a preference yes or no?
@X-sam  @jonathanperret

jonathanperret: My vote is still for no preference.

mathPi: I would also vote for no preference as I don't think someone would want his computer to sleep while knitting.

dl1com: Ok, then let's go without a preference for now. If somebody finds it really annoying we can add the preference setting later on.
@mathPi would you mind adjusting your PR?

mathPi: It is already done ;)

jonathanperret:
Here's a test release from this PR: https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.1-keepawake

I think it would be nice to have confirmation that this PR works on all supported platforms before merging. I just test it on macOS and Linux and can confirm it works on those.

dl1com:
> Here's a test release from this PR: https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.1-keepawake
> 
> I think it would be nice to have confirmation that this PR works on all supported platforms before merging. I just test it on macOS and Linux and can confirm it works on those.

@clholgat @Adrienne200 @t0mpr1c3 could you verify this works on Windows by any chance?

jonathanperret:
> > Here's a test release from this PR: [jonathanperret/ayab-desktop@`1.0.1-keepawake` (release)](https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.1-keepawake)
> > I think it would be nice to have confirmation that this PR works on all supported platforms before merging. I just test it on macOS and Linux and can confirm it works on those.
> 
> @clholgat @Adrienne200 @t0mpr1c3 could you verify this works on Windows by any chance?

I just borrowed a Windows laptop, I'll be checking it right now with the latest build (https://github.com/AllYarnsAreBeautiful/ayab-desktop/actions/runs/14431786684?pr=699)

jonathanperret:
The Windows test is successful: the computer no longer goes to sleep during knitting (I tested with simulation). I re-confirmed that with the `main` version of AYAB it would go to sleep while knitting. I also checked that sleep was only prevented during knitting and not when the app was idle.

This looks ready to go in!

jonathanperret (approved):

dl1com (approved):

Issue #678, opened 2024-07-21
The machine type displayed is not updated when changed in Preferences


jonathanperret:
**Environment**
AYAB software version: 1.0.0-dev (bc96d5c3e3f955fa75551ec44c82e966e0981112)
Computer/OS: macOS 14.2

**Describe the bug**
The machine type displayed is not updated when changed in Preferences.

**To Reproduce**
1. Launch AYAB Desktop
2. Click on Preferences/Set Preferences
3. Change the machine type
4. Click OK
5. See that the machine type in the right bar is unchanged

**Expected behavior**
The machine type should be updated.

**Screenshots**

https://github.com/user-attachments/assets/3e00450c-4bf9-4732-9b07-4c275da57120

mathPi:
The issue seems even more important than just the machine name not updated. I think we should trigger a config refresh because the needle number can very when switching machines.

jonathanperret:
> The issue seems even more important than just the machine name not updated. I think we should trigger a config refresh because the needle number can very when switching machines.

This is correct, but the needle count is already being updated when the machine type changes. As far as I can tell, only the machine type displayed fails to be updated.

mathPi:
Ohh I didn't find the needle count update in the code.

PR #700, opened 2024-08-25
fix: #678 refresh the engine configuration if settings change

mathPi:
To fix issue #678, we need to refresh the engine parameters as the new settings can impact those parameters.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for refreshing window preferences, enhancing user interface responsiveness.
	- Added a validation method that improves form handling and updates the main GUI upon user interaction.
	- Implemented a method to close dialogs while ensuring UI refresh, providing immediate feedback on preference changes.

- **Improvements**
	- Enhanced modularity in the GUI component initialization process for better maintainability and consistency of displayed information.

### DIFF
--- a/src/main/python/main/ayab/engine/engine.py
+++ b/src/main/python/main/ayab/engine/engine.py
@@ -61,7 +61,7 @@ class Engine(SignalSender, QDockWidget):
         self.ui.setupUi(self)
         self.config: OptionsTab = OptionsTab(parent)
         self.config.portname = self.__read_portname()
-        self.config.refresh()
+        self.reload_settings()
         self.status = StatusTab()
         self.setup_ui()
         parent.ui.dock_container_layout.addWidget(self)
@@ -70,7 +70,6 @@ class Engine(SignalSender, QDockWidget):
         self.control = Control(parent, self)
         self.__feedback = FeedbackHandler(parent)
         self.__logger = logging.getLogger(type(self).__name__)
-        self.refresh_settings_change()
 
     def __del__(self) -> None:
         self.control.stop()
@@ -90,7 +89,7 @@ class Engine(SignalSender, QDockWidget):
         # activate UI elements
         self.__activate_ui()
 
-    def refresh_settings_change(self) -> None:
+    def reload_settings(self) -> None:
         self.config.refresh()
         self.setWindowTitle("Machine: " + Machine(self.config.machine).name)
 
--- a/src/main/python/main/ayab/engine/engine.py
+++ b/src/main/python/main/ayab/engine/engine.py
@@ -70,7 +70,7 @@ class Engine(SignalSender, QDockWidget):
         self.control = Control(parent, self)
         self.__feedback = FeedbackHandler(parent)
         self.__logger = logging.getLogger(type(self).__name__)
-        self.setWindowTitle("Machine: " + Machine(self.config.machine).name)
+        self.refresh_settings_change()
 
     def __del__(self) -> None:
         self.control.stop()
@@ -89,6 +89,10 @@ class Engine(SignalSender, QDockWidget):
 
         # activate UI elements
         self.__activate_ui()
+
+    def refresh_settings_change(self) -> None:
+        self.config.refresh()
+        self.setWindowTitle("Machine: " + Machine(self.config.machine).name)
 
     def __disable_status_tab(self) -> None:
         self.ui.tab_widget.setTabEnabled(1, False)
--- a/src/main/python/main/ayab/ayab.py
+++ b/src/main/python/main/ayab/ayab.py
@@ -129,7 +129,7 @@ class GuiMain(QMainWindow):
     def __set_prefs(self) -> None:
         self.prefs.open_dialog()
         self.scene.refresh()
-        self.engine.refresh_settings_change()
+        self.engine.reload_settings()
 
     def __quit(self) -> None:
         logging.debug("Quitting")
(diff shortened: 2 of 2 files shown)


src/main/python/main/ayab/preferences.py
```diff
@@ -226,7 +226,7 @@ def __init__(self, parent: GuiMain):
         for widget in self.__widget.values():
             widget.connectChange()
         self.__ui.reset.clicked.connect(self.__reset_and_refresh)
-        self.__ui.enter.clicked.connect(self.accept)
+        self.__ui.enter.clicked.connect(self.__validate_form)
```
jonathanperret, line 229: The way the Preferences dialog is currently setup, the `OK` button is a lie. Each setting is updated as soon as you change its value, event though nothing happens immediately in the main window because the main event loop is blocked by the popup. You can verify this by closing the Preferences dialog without clicking `OK`, i.e. press Escape or click the window close button. Demonstration:

https://github.com/user-attachments/assets/970f7fde-77dd-4134-9bc1-0e202594be91

Therefore, intercepting the click on the `OK` button would make the UI refresh only for users that click it, causing weird behaviors.

The Preferences dialog should probably be reworked to actually commit settings only when `OK` is clicked, but that's a bigger piece of work.

mathPi: Ohhh yes, I missed that :( I'll try to find an other solution.

mathPi: @jonathanperret the done method seems to be triggered in all cases that allow to close the window, (ESC, close button, OK button). Do you see something else ?

jonathanperret:
There are already two different places (that I could find!) where the UI is refreshed following a preferences update:

1. In `preferences.py`, right after the (blocking) call that opens the Preferences dialog, an `image_resizer` signal is (sometimes) emitted: https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/a03774e6a62410f3c4e57394d949b8b803860ac4/src/main/python/main/ayab/preferences.py#L194-L198
2. In `ayab.py`, there's a call to `scene.refresh()` after the (blocking) call to the above method: https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/a03774e6a62410f3c4e57394d949b8b803860ac4/src/main/python/main/ayab/ayab.py#L129-L131

I think it would be good not to add a third place to that list…

mathPi:
> There are already two different places (that I could find!) where the UI is refreshed following a preferences update:
> 
> 1. In `preferences.py`, right after the (blocking) call that opens the Preferences dialog, an `image_resizer` signal is (sometimes) emitted: https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/a03774e6a62410f3c4e57394d949b8b803860ac4/src/main/python/main/ayab/preferences.py#L194-L198
> 2. In `ayab.py`, there's a call to `scene.refresh()` after the (blocking) call to the above method: https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/a03774e6a62410f3c4e57394d949b8b803860ac4/src/main/python/main/ayab/ayab.py#L129-L131
> 
> I think it would be good not to add a third place to that list…

Ohh yes, the second one is perfect, I update the PR. Is it better now ?

src/main/python/main/ayab/engine/engine.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, parent: GuiMain):
         self.control = Control(parent, self)
         self.__feedback = FeedbackHandler(parent)
         self.__logger = logging.getLogger(type(self).__name__)
-        self.setWindowTitle("Machine: " + Machine(self.config.machine).name)
+        self.refresh_settings_change()
```
jonathanperret, line 73: This causes `self.config.refresh()` to be called twice in `__init__`, maybe this could be improved?

jonathanperret:
> Ohh yes, the second one is perfect, I update the PR. Is it better now ?

Yes, that's definitely better (less code, yay). Although the Qt way would be to use a signal so that things are less coupled (e.g. `Engine` could listen to a `preferences_changed` signal), but that's not how the code is structured now unfortunately, so that's another change that will have to wait.

There is a change in behavior introduced by this though. The `Engine.refresh()` method is slightly misnamed as it actually resets (most) options to their default values. With the previous code, the `Infinite repeat` checkbox for example was not reset to the default when closing the Preferences dialog. This is probably not a big deal, just maybe a bit surprising.
In any case, the same options are also reset when opening a new image, which I find annoying by the way — sometimes you just want to reload the image that you've modified in an editor — so this new behavior wouldn't be so much out of line with the rest.

I have one last nit to pick with the name of that new method: I'm not sure `refresh_settings_change` makes sense as a phrase, but maybe I'll let a native English speaker give their opinion on that one. I'd have gone with `reload_settings` maybe?

Finally, thank you for taking my feedback into account but I don't have permission to approve PRs in this repo, you'll have to wait for a maintainer to drop by.

src/main/python/main/ayab/preferences.py
jonathanperret: For anyone reading this later, [my later comment](https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/700#issuecomment-2308759538) and @mathPi 's following changes obviate this discussion.

dl1com:
> I have one last nit to pick with the name of that new method: I'm not sure `refresh_settings_change` makes sense as a phrase, but maybe I'll let a native English speaker give their opinion on that one. I'd have gone with `reload_settings` maybe?

Waiting for clarification on this, then merge

dl1com (approved):

PR #708, opened 2024-09-14
issue template: put instructions in comments so they don't appear in the issue

jonathanperret:
## 🤔 Problem
Having an issue template is great, but people don't always think to remove the instructions (such as `A clear and concise description of what you expected to happen.`) when filing an issue.

This results in slightly confusing reports where the instructions appear together with the actual issue content. 

## 💡 Proposed solution

Putting the instructions in markdown comment blocks at least prevents them from visually polluting the resulting issue.

## Summary by CodeRabbit

- **Documentation**
	- Updated the bug report template to use HTML comments for guidance, enhancing user experience by clarifying instructions without cluttering the visible content.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,7 +14,7 @@ Knitting machine:
 AYAB hardware: 
 
 **Describe the bug**
-A clear and concise description of what the bug is.
+<!-- A clear and concise description of what the bug is. -->
 
 **To Reproduce**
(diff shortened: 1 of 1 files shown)


dl1com:
Let's see how it goes, there might be users putting their text inside the comment now....but let's give it a try

dl1com (approved):

jonathanperret:
> Let's see how it goes, there might be users putting their text inside the comment now....but let's give it a try

Yes, this would be the risk. We’ll see.

PR #712, opened 2024-09-22
Fix "default knit side image" behavior

### LINKED ISSUE #711: [BUG] "Default Knit Side Image" does not flip loaded images
**Environment**
AYAB software version: 1.0-beta1
Computer/OS: macOS Sonoma 14.6.1
Knitting machine: KH-910
AYAB hardware: AYAB shield

**Describe the bug**
With the "Default Knit Side Image" preference checked, when an image is loaded, it should be horizontally flipped so that it appears correct on the knit (right) side. Currently, it does not get flipped.

**To Reproduce**
1. Open the **Preferences** dialog, ensure **Default Knit Side Image** is checked, close the dialog;
2. Click **Open Image File** and select `uc3_30x15.png`;
3. Observe that in the Settings pane, **Knit Side Image** is checked;
4. Observe that in the main panel, the image is not flipped (`µC3` is readable);
5. Knit or simulate knitting: observe that the image is also knitted unflipped, so that:
   * the knit progress pane shows the image's left pixels (starting with the stem of the `µ`) being knitted on the left;
   * `µC3` is readable on the purl side, not the knit side.

image

**Expected behavior**
With the **Default Knit Side Image** preference is checked, an image should be flipped as it is loaded.

Note that if **Default Knit Side Image** is left unchecked, and the user checks **Knit Side Image** after loading the image, it correctly gets flipped:

image

**Additional context**

In the [1.0 release notes](https://github.com/AllYarnsAreBeautiful/ayab-manual/pull/47), we suggest that users who want the historical AYAB behavior of always flipping loaded images can check **Default Knit Side Image**, so it's not unlikely that a number of users would hit this bug as they migrate to 1.0.

jonathanperret:
## Problem

Fixes #711.

## Proposed solution

Instead of `Scene` modifying `AyabImage`'s internal `image`, it now reverses the image (if requested) only at render time. `Pattern` does likewise when generating the pattern from the image.

This makes it easier to avoid getting confused with how many times the image has been reversed.


src/main/python/main/ayab/engine/options.py
```diff
@@ -122,14 +123,15 @@ def update_needles(self) -> None:
         stop_needle = NeedleColor.read_stop_needle(self.ui, self.machine)
         self.emit_needles_updater(start_needle, stop_needle)
 
-    def __reverse_image(self) -> None:
-        if self.ui.auto_mirror_checkbox.isChecked():
+    def __auto_mirror_changed(self) -> None:
+        reversed = self.ui.auto_mirror_checkbox.isChecked()
```
dl1com, line 127: Would you mind to use another name for this variable, as the current name is already taken by a Python builtin method? (see SonarQube analysis)

jonathanperret: Ooh, good catch. I'll remember to check the sonar analysis in the future.

I've gone ahead and renamed all occurrences of `reversed` to `image_reversed`.

dl1com (approved):

PR #718, opened 2024-09-30
DAK file reading: don't assume UTF-8 encoding

### LINKED ISSUE #716: [BUG] Error when loading .stp files with non-ASCII colors
**Environment**
AYAB software version: 1.0.0-beta1
Computer/OS: Mac

**Describe the bug**
When trying to open a stp file from DAK8, I have an error message "Unable to load the image file".

Here is an example stp file:
[fond.stp.zip](https://github.com/user-attachments/files/17154249/fond.stp.zip)

jonathanperret:
## Problem

Fixes #716.

## Proposed solution

When extracting strings from a DAK (.stp or .pat) format file, we shouldn't assume that these strings are encoded using UTF-8.

In particular, it appears localized color names are not encoded using UTF-8, and trying to decode them as UTF-8 causes pattern loading to fail.

Another place where strings are read from the files is as part of the .stp decryption process. In that case, it seems safer to not try to interpret those byte sequences at all.

This PR changes the strings to Python bytestrings, so we don't need to guess at the encoding that DAK actually uses for color names, and we can be sure the bytes in the encryption keys are unmodified.

## How to test

Try to load the file attached in #716.

## Summary by CodeRabbit

- **New Features**
	- Enhanced data handling by updating string types to bytes for improved compatibility and performance.
  
- **Bug Fixes**
	- Adjusted methods to ensure proper functioning with the new byte data types, enhancing reliability.

- **Refactor**
	- Improved code readability and consistency through minor formatting adjustments.

### DIFF
--- a/src/main/python/main/ayab/pattern_import.py
+++ b/src/main/python/main/ayab/pattern_import.py
@@ -32,9 +32,9 @@ def getDWordAt(data: bytes, i: int) -> np.uint32:
 
 
 # Pascal-style string
-def getStringAt(data: bytes, i: int) -> str:
+def getStringAt(data: bytes, i: int) -> bytes:
     size = getByteAt(data, i)
-    return data[i + 1 : i + size + 1].decode()
+    return data[i + 1 : i + size + 1]
 
 
 class Color:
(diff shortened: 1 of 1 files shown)


mathPi:
I tested using one of my files and running a simulation and it is working.

dl1com (approved):

dl1com:
> I tested using one of my files and running a simulation and it is working.

Thanks for your feedback.
Plus the additional feedback in Discord, this looks good for now and ready to merge.

PR #722, opened 2024-10-29
Fix Linux build after python-appimage release disappeared

### LINKED ISSUE #721: [BUG] Linux build is broken because of missing `python-appimage` dependency
I hit this accidentally building a test release on my fork but it will happen next time we try to build a release here: https://github.com/jonathanperret/ayab-desktop/actions/runs/11542883130/job/32126632072

It appears the `https://github.com/niess/python-appimage/releases/download/python3.11/python3.11.9-cp311-cp311-manylinux_2_28_x86_64.AppImage` link is no longer valid. Somehow the https://github.com/niess/python-appimage project is pulling artifacts from their releases.

jonathanperret:
## Problem

Fixes #721 — at least until `python-appimage` breaks us again.

## Proposed solution

This PR slightly changes the logic used to download the `python-appimage` build: instead of a hardcoded `.9` added to the specified Python version (`3.11`), the build now uses a full Python version (e.g. `3.11.10`) and tries to download a matching `python-appimage` build.

Note that the mismatch between the Python versions used for Windows and macOS (`3.11.9`) on one side, and Linux (`3.11.10`) on the other, is caused by the fact that:
 * a `python-appimage` build is now only available for Python `3.11.10`;
 * `actions/setup-python` cannot install Python `3.11.10` on Windows or macOS (apparently because Python 3.11 is EOL, there are no longer official binaries distributions for the security updates).

Note that because we previously were not specifying the full Python version, these specific versions were already being used by the build.

I'd like to investigate the https://github.com/niess/python-appimage release process a bit more to understand why they remove previously published assets, but for the immediate term this fix should let us continue building AYAB releases.

## Summary by CodeRabbit

- **New Features**
	- Enhanced clarity in Python version handling within GitHub Actions workflows.
  
- **Bug Fixes**
	- Updated Python versions in multiple workflows to more specific minor versions, improving consistency and reliability in builds and tests.

- **Chores**
	- Restructured workflow steps for clearer execution order and improved output variable naming conventions.

### DIFF
--- a/.github/actions/init-environment/action.yml
+++ b/.github/actions/init-environment/action.yml
@@ -8,13 +8,19 @@ outputs:
     value: ${{steps.vars.outputs.draft}}
   manifest:
     value: ${{steps.vars.outputs.manifest}}
-  python:
-    value: ${{steps.vars.outputs.python}}
+  python-minor:
+    value: ${{steps.vars.outputs.python-minor}}
   python-appimage:
     value: ${{steps.vars.outputs.python-appimage}}
 runs:
   using: "composite"
   steps:
+    - name: Set up Python ${{ matrix.python-version }}
+      id: py
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+        cache: "pip"
     - name: Set variables
       id: vars
       shell: bash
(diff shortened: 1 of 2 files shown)


dl1com (approved):

dl1com:
Thanks for taking care of this...supporting multiple OS is hard :(

Issue #662, opened 2024-04-21
Patterning can get stuck on row n-1 and never finish


Adrienne200:
AYAB software version: 0.99.0-rc6
Computer/OS: Intel Mac, Big Sur
Knitting machine: KH910
AYAB hardware: Interface

 It's new in the rc4-rc6 timeframe, I've never seen this happen before.

Steps:
All settings and prefs at default, so Infinite Repeat is unchecked
You need to actually air-knit, Simulate is different and doesn't show the issue.
Open the Triangles sample. It is 10 rows.
(Optional: set the start row to 7 for more convenient testing.)
Start on the left, remembering to pause for two seconds after crossing the turn mark.
and knit as usual for 8 rows, noticing the end-of-row sounds from the hardware's buzzer and the "ding" from computer's speakers.
Keep going for rows 9 and 10 without slowing down.

Result: Usually it will select row 10 and complete as expected. But sometimes (about half the time on Mac, rarely on Windows but it did happen) I could get it to just keep repeating row 9 without ever finishing.

X-sam:
possibly related to AllYarnsAreBeautiful/ayab-firmware#190 ?

jonathanperret:
I've done a number of swatches with (recent) 1.0-dev firmware and not hit this, I think. @Adrienne200 are you up for trying it with `1.0.0-beta1`?

jonathanperret:
Thanks for testing again. I’ll give it a try tomorrow on my 910 (from a Mac) with your exact instructions to see if I can reproduce it.

jonathanperret:
@Adrienne200 I'm trying to looking into this more seriously. I seem to remember on the call last week you sais it occurred only with an AYAB Interface?
One thing I'd like to understand more clearly is what you see happening on the last row. Does the desktop app ever get to the point where it says "Finished knitting"? Or does it just stop reacting to rows being completed on the machine? Or does it keep emitting "dings" but repeating row 9?

clholgat:
I can easily repro on my 270 with a Jens interface.

If the last row is row N. I never get the end of row beep for row N-1 and then a few seconds later, the desktop software gives me the end of pattern beeps but nothing more from the interface.

Digging.

clholgat:
Found a fix https://github.com/AllYarnsAreBeautiful/ayab-firmware/pull/217

jonathanperret:
🛎 A proposed fix for this is in #740 and can be tested in the release referenced there.

PR #731, opened 2024-12-03
Synchronise end-of-line sounds between app and firmware

### LINKED ISSUE #617: [BUG] Incorrect sounds/display at the end of a single repeat
AYAB software version: 0.99 RC2
Computer/OS: Windows 7
Knitting machine: KH910
AYAB hardware: Interface or shield

When the Infinite Repeat box is unchecked, there's a specific sequence to guide the user through the last couple of rows. The current functionality is correct, but the associated sounds are not, they don't match up to the text prompts. Either the sounds or the text need to change. 
When I first looked at this I thought it was the long sound that had gone missing, but on deeper investigation there is a new sound that I maybe like better, it just needs to be a bit beefier and have a text change to match the new sequence.

1. Open the Triangles sample. It is 10 rows.
2. All settings on default, the important one is that Infinite Repeat is unchecked.
3. Start on the left and knit as usual for 8 rows, noticing the end-of-row sounds from the hardware's buzzer and the "ding" from computer's speakers.
4. Slow down for rows 9 and 10 and observe carefully.

Expected: (For K and L carriage. G may be different.)
Take the carriage across for row 9, and as you turn around there's the usual EOR sounds, [Edit: no, the EOR sound from the computer is different.] and the display changes to Row 10 as usual. (It is about to do the selection for row 10, so you'll need to actually knit two more passes for Fair Isle.)
This is the place to show a warning that you're almost finished, and a warning sound. (It does this)
The Row 10 text should remain as you bring the carriage across for row 10.
Now there's no more selection to do, but you need to knit "Row 11". Row 11 should unselect all needles to B.
If we still want the double-beep sound, it would go here at the end of row 10 (beginning of row 11) consistent with warning you as you turn around. But it’s only one pass after the previous sound.
Alternatively, now that there's a different warning sound from the computer as you turn between 9 and 10, I'd be OK with doing away with the old double-beep warning sound (and changing the text accordingly.) Beef up the sound that now happens as the warning message is posted, possibly repeat it in the hardware buzzer, and rewrite its text.

Actual: 
Observing at the turnaround between 9 and 10:
- there's the usual EOR sound from the buzzer (OK)
- no EOR sound from the computer (this is OK)
- the text for "Row 10/10" flashes briefly, then it disappears
- the "Image Transmission finished. Knit until you hear the double-beep sound" warning appears
- along with a different sound from the computer. 
I'm OK with this so far. This is the correct time for this to happen, and having the computer do a new sound here is good. Maybe the buzzer should sound as well? 
But it would be nice if Row 10/10 could continue to be displayed as you knit across, as in previous rows.
Now bring the carriage across to select row 10 and stop at the end:
- the selection has happened as usual (OK)
- there's an EOR sound from the buzzer (OK)
- but no sound from the computer. (The long double-beep used to happen here. If we remove the double-beep it should be a single beep instead.)
Take the carriage across to knit "row 11"
 - it selects all needle to B. (Correct)
 - no sounds from either the buzzer or the computer. (Expected, it has finished its job at the end of row 10.)

For reference:
How did Brother handle this? KH930 native does this:
TL;DR: In a 10-row pattern a warning sound happens as you turn around after row 9 and are about to select row 10.
In detail:
As with AYAB, the row numbers on the display are those of the selection that is happening, ignoring the fact that the actual knitting is a row behind.
There is no single repeat function as such, it will always select row 1 of the next repeat; the user has to know to not knit pattern row 1 again even though it has selected; but it does have a system to warn you.
There's no end-of-row sound, but the row number counts up on the display, solid while the carriage moves across, flashing the old number after you've passed the middle of the bed but haven't turned around yet, steps up to the next number (solid) as you turn. 
There is an end-of-repeat sound, which is how it warns you that you need to take action to not knit row 1 again if you don't want to. In a 10-row pattern that sound happens as you turn around after row 9 and are about to select for row 10. 
You will normally knit two more passes after the beep, one to select for row 10, then row 11 to actually knit the row 10 selection, which will also select row 1 again. (When knitting fast, you notice this beep while doing the row 10 selection row, so it feels like just one row of warning, which is what you want; warning a row later would sometimes be too late.)
For row 12, you switch the KC knob to NL so that it will "gather up" all the needles and treat them as if they were at B, even though they have pre-selected for row 1 while you were doing row 11 which knits the selection from row 10.

jonathanperret:
## 🤔 Problem

The end-of-line sounds are not synchronized between the firmware and the desktop app, fixes #617.

## 💡 Proposed solution

The main source of discrepancy is the fact that the desktop app closes the connection to the firmware as soon as it has sent the final pattern row. But this actually happens two full carriage passes (or one, with the garter carriage) before knitting is really complete.

The firmware itself continues to emit end-of-row sounds until the end of the last selecting row, i.e. one pass further than the desktop app.

This PR slightly changes the final row transmission logic: instead of marking the last patterning row as "final" and closing the connection, it sends that last row normally (not telling the firmware that it is the last), so that it gets the usual request for a next row from the firmware at the end of that pass, which the app now answers with a blank row.

This which lets the app emit a "work finished" sound at the same point that previous AYAB versions emitted an "end of work" beep from the hardware.

This also makes the message "Please knit until you hear the double beep sound" notification more correct, as the final pass (knitting the final selection of needles while moving them all back to B) will trigger one last sound from the hardware.

Two other minor issues around sounds are also fixed by this PR:
 - the "work finished" sound from the desktop app, in addition to being one pass early as described above, was delayed by one to two seconds — it is now synchronous with the hardware beep, making the app feel more "snappy";
 - the "simulation" mode only emitted end-of-line sounds for the first row, it now more faithfully reproduces the experience of actual knitting.

## 🤔 Remarks

This change may also "fix" #662, or at least reduce its impact — if the firmware now fails to "move on" from the final blank row, the result will be very close to having properly finished the process.

Also, the hardware sounds themselves need a bit of work, as for example the final "double beep sound" does not actually sound like a double beep, at least on the AYAB shield's standard buzzer. I have changes to the firmware that should improve that situation, which I will submit in an upcoming PR to `ayab-firmware`.

## 🔬 How to test

I have prepared a test release at https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.0-endsounds-4 that can be installed as usual. It does not include a new firmware (compared to 1.0.0 beta 3) so no reflashing should be necessary.

Note that this PR (and the test release above) now includes PR #740 which is a proper fix for #662, so neither it nor #617 should be reproducible on the test release.

## Release Notes

- **New Features**
	- Introduced a new `FINISHING` state in the state machine to enhance the knitting operation's control flow.
	- Added audio feedback for knitting completion.

- **Improvements**
	- Simplified signal handling by removing unnecessary parameters from various methods.
	- Enhanced the testing framework for better validation of communication functionality.
	- Improved the efficiency of the knitting operation's completion process.

- **Bug Fixes**
	- Streamlined the operation completion process by removing redundant audio notifications.

### DIFF
--- a/src/main/python/main/ayab/ayab.py
+++ b/src/main/python/main/ayab/ayab.py
@@ -106,9 +106,7 @@ class GuiMain(QMainWindow):
             self.scene.ayabimage.select_file
         )
         self.ui.cancel_button.clicked.connect(self.engine.cancel)
-        self.hw_test.finished.connect(
-            lambda: self.finish_operation(Operation.TEST, False)
-        )
+        self.hw_test.finished.connect(lambda: self.finish_operation(Operation.TEST))
 
     def __activate_menu(self) -> None:
         self.menu.ui.action_open_image_file.triggered.connect(
@@ -158,7 +156,7 @@ class GuiMain(QMainWindow):
         self.ui.open_image_file_button.setEnabled(False)
         self.menu.setEnabled(False)
 
-    def finish_operation(self, operation: Operation, beep: bool) -> None:
+    def finish_operation(self, operation: Operation) -> None:
         """(Re-)enable UI elements after operation finishes."""
         if operation == Operation.KNIT:
             self.knit_thread.wait()
@@ -168,9 +166,6 @@ class GuiMain(QMainWindow):
         self.ui.filename_lineedit.setEnabled(True)
         self.ui.open_image_file_button.setEnabled(True)
         self.menu.setEnabled(True)
-
-        if operation == Operation.KNIT and beep:
-            self.audio.play("finish")
 
     def set_image_dimensions(self) -> None:
         """Set dimensions of image."""
--- a/src/main/python/main/ayab/engine/communication_mock.py
+++ b/src/main/python/main/ayab/engine/communication_mock.py
@@ -23,20 +23,17 @@ Mock Class of Communication for Test/Simulation purposes
 import logging
 from time import sleep
 
-from PySide6.QtWidgets import QMessageBox
-
 from .communication import Communication, Token
 
 
 class CommunicationMock(Communication):
     """Class Handling the mock communication protocol."""
 
-    def __init__(self, delay=True, step=False) -> None:
+    def __init__(self, delay=True) -> None:
         """Initialize communication."""
         logging.basicConfig(level=logging.DEBUG)
         self.logger = logging.getLogger(type(self).__name__)
         self.__delay = delay
-        self.__step = step
         self.reset()
 
     def __del__(self) -> None:
@@ -122,20 +119,6 @@ class CommunicationMock(Communication):
                 self.rx_msg_list.append(reqLine)
                 if self.__delay:
                     sleep(1)  # wait for knitting progress dialog to update
-                # step through output line by line
-                if self.__step:
-                    # pop up box waits for user input before moving on to next line
-                    msg = QMessageBox()
-                    msg.setIcon(QMessageBox.Icon.Information)
-                    msg.setText("Line number = " + str(self.__line_count))
-                    msg.setStandardButtons(
-                        QMessageBox.StandardButton.Ok
-                        | QMessageBox.StandardButton.Cancel
-                    )
-                    ret = None
-                    ret = msg.exec_()
-                    while ret is None:
-                        pass
             else:
                 self.__started_row = True
         if len(self.rx_msg_list) > 0:
--- a/src/main/python/main/ayab/engine/communication_mock.py
+++ b/src/main/python/main/ayab/engine/communication_mock.py
@@ -22,7 +22,6 @@ Mock Class of Communication for Test/Simulation purposes
 
 import logging
 from time import sleep
-from collections import deque
 
 from PySide6.QtWidgets import QMessageBox
 
@@ -47,8 +46,9 @@ class CommunicationMock(Communication):
     def reset(self):
         self.__is_open = False
         self.__is_started = False
-        self.rx_msg_list = deque([], maxlen=100)
+        self.rx_msg_list = list()
         self.__line_count = 0
+        self.__started_row = False
 
     def is_open(self) -> bool:
         """Return status of the interface."""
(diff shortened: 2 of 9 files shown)


jonathanperret:
> @jonathanperret do we require an update in the API description to reflect your changes?

I don't think so, I am using the API exactly as it is documented — as demonstrated by the fact that the firmware doesn't need to be modified for this code to work.

jonathanperret:
I just triggered a new test release that now includes the newly merged fixes for #732 and #734 : https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.0-endsounds-3

jpcornil-git:
Note that with this change, the FW doesn't move to init state anymore unless you effectively knit the blank row.

jonathanperret:
> Note that with this change, the FW doesn't move to init state anymore unless you effectively knit the blank row.

Note that in general you do have to "knit" the "blank row", because while it will select every needle to `B`, it will knit the final row of the pattern, that was previously selected. The only exception is the garter carriage, whose work is indeed finished as soon as it reaches the end of the last pattern row.

In theory there can be a difference between the firmware being in `init` state and still executing a blank row, but in practice it is not easily observable. Remember that the desktop app always closes the serial port after knitting, and opening it back up resets the Arduino so the question of the final state is moot.

By the way, the current firmware would not work if that reset did not occur (you can easily see this with `simavr`), because even if you knit until it goes back to `init` state, it does not accept the initial `reqInit` message from the app (that message is only accepted in `wait_for_machine` state).

This is all to say that yes, the current firmware and API are a bit broken, and will have to be improved, but I am trying to make a small change here for now.

jpcornil-git:
That reset is specific to the current AVR board, if you take a uno R4 minima or Wifi for example, then opening the serial port doesn't reset the device, i.e. behavior is not generic but specific to a few platform => why in ayabAsync, I do accept an init message while in operation mode to support R4, see
https://github.com/jpcornil-git/ayabAsync-firmware/blob/v1.0/lib/components/knitter/knitter.cpp#L128

It will be required as well moving forward with a webapp/wifi communication (no HW reset).

From an user experience, the HW/shield beep will also be missing (unless you knit the blank row again).

jonathanperret:
> That reset is specific to the current AVR board, if you take a uno R4 minima or Wifi for example, then opening the serial port doesn't reset the device, i.e. behavior is not generic but specific to a few platforms => why in ayabAsync, I do accept an init message while in operation mode to support R4, see [jpcornil-git/ayabAsync-firmware@`v1.0`/lib/components/knitter/knitter.cpp#L128](https://github.com/jpcornil-git/ayabAsync-firmware/blob/v1.0/lib/components/knitter/knitter.cpp?rgh-link-date=2024-12-04T18%3A38%3A40Z#L128)
> 
> It will be required as well moving forward with a webapp/wifi communication (no HW reset).

Again, I'm 100% onboard with improving the firmware and API in the future.

> From an user experience, the HW/shield beep will also be missing (unless you knit the blank row again).

The beep is here. Have you tried knitting with `ayab-firmware`?

jpcornil-git:
No (remember that I've no machine/only coding for my daugther :-)) but I see that Knitter::stopKnitting() and therefore GlobalBeeper::endWork() won't be called until you finish the last line
https://github.com/AllYarnsAreBeautiful/ayab-firmware/blob/main/src/ayab/knitter.cpp#L454
```c
void Knitter::stopKnitting() const {
  GlobalBeeper::endWork();
  GlobalFsm::setState(OpState::init);
```

jonathanperret:
> No (remember that I've no machine/only coding for my daugther :-)) but I see that Knitter::stopKnitting() and therefore GlobalBeeper::endWork() won't be called until you finish the last line

Yes, but `GlobalBeeper::finishedLine` is called after every row is received. Also it may look like `GlobalBeeper::endWork` would be make a special sound but 1) the sounds are all borked up on the shield beeper due to switching too fast for the electromechanical device and 2) if you look closer you'll see that currently there's another `GlobalBeeper::finishedLine` call right after the `GlobalBeeper::endWork`, which resets the beep count and nullifies the `GlobalBeeper::endWork`.

Believe me, I know that so many things are suboptimal in the whole thing. I'm just trying to start somewhere and improve things bit by bit here. I believe the change in this PR makes things incrementally better. Let's remember the focus here is to get to a state where we can make the first release of AYAB in over 5 years.

jpcornil-git:
I understand but I believe that adding the suggested indState is a step in the right direction while adding a fake new row is rather a workaround/a step in the wrong direction that you will have to undo moving forward.
I believe you got my points/won't insist anymore on this :-)

dl1com (approved):

dl1com:
Reported to be working by @Adrienne200 in Discord

PR #739, opened 2024-12-05
bool -> bit for req_init flags

clholgat:
https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/662

## Summary by CodeRabbit

- **New Features**
	- Enhanced the communication method to transmit flags as single bits, improving data efficiency.
  
- **Documentation**
	- Updated comments to clarify the expectation of bit representation for flags in the API.

### DIFF
--- a/src/main/python/main/ayab/engine/communication.py
+++ b/src/main/python/main/ayab/engine/communication.py
@@ -186,7 +186,10 @@ class Communication(object):
         data.append(Token.cnfLine.value)
         data.append(line_number)
         data.append(color)
-        data.append(flags)
+
+        # Python ints are at least a byte
+        # The API wants bits
+        data.append(0b1 if flags else 0b0)
         data.extend(line_data)
         hash = 0
         hash = add_crc(hash, data)


jpcornil-git:
Not sure why this is required, `boolean` are (a subclass of) `int` already
```python
>>> help(bool)
Help on class bool in module builtins:

class bool(int)
 |  bool(x) -> bool
 |  
 |  Returns True when the argument x is true, False otherwise.
 |  The builtins True and False are the only two instances of the class bool.
 |  The class bool is a subclass of the class int, and cannot be subclassed.
...
```
=> proposed change is replacing an integer by an integer of the same size (0b is only a formater)

If you really want to control the conversion to byte, you should use something like
```
data.extend(flag.to_bytes(byteorder='little', length=1))

```
... but this is not required as it is done already within the `append` method of the `bytearray` object that takes an `int` as input.

What issue # are we trying to solve here btw ?

jonathanperret:
I made a release from this PR if anyone wants to test: https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.0-never-end-1

I couldn't reproduce #662 and I still can't with this build, but it does not seem to have broken anything either.

dl1com:
Issue #662 fixed by #740

PR #741, opened 2024-12-11
Update ayab-translation-master.tsv

Netclicks:
Added Japanese translations to the translation master tsv file. I am still missing a few translations, specifically for 'tools' and the terms relating to 'halls'


dl1com:
@Netclicks Thanks a lot for the contribution! We'll check if it breaks something (i.e. encoding issues) and decide whether it will get into 1.0 or the next version.

dl1com (approved):
tsv seems to be valid

PR #742, opened 2024-12-13
Linux build: bump to Python 3.11.11

jonathanperret:
## Problem

[python-appimage](https://github.com/niess/python-appimage) removed the 3.11.10 release from their repo.

## Proposed solution

Like we did [last time](https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/722), we bump the version of Python used for the Linux build.

This is not satisfying but right now I don't have a better idea.

## Summary by CodeRabbit

- **New Features**
	- Enhanced AppImage creation process with new steps for downloading, extracting, and installing dependencies.
  
- **Improvements**
	- Updated Python version for Linux builds to 3.11.11.
	- Standardized caching strategies across all build jobs for consistency.

- **Documentation**
	- Added comments to improve clarity and maintainability of the workflow.

### DIFF
--- a/.github/workflows/build-multi-os.yml
+++ b/.github/workflows/build-multi-os.yml
@@ -440,8 +440,8 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        # Using Python 3.11.10 specifically for Linux builds due to python-appimage availability
-        python-version: [3.11.10]
+        # Using Python 3.11.11 specifically for Linux builds due to python-appimage availability
+        python-version: [3.11.11]
     steps:
       - name: Checkout repo into AppDir
         uses: actions/checkout@v4


jonathanperret:
> .github/workflows/build-multi-os.yml (1)> `443-444`: **Consider future alignment of Python versions across platforms.**
> > While the current patch-level differences between platforms (3.11.9 for Windows/macOS vs 3.11.11 for Linux) should maintain compatibility, consider aligning these versions in a future update if possible. This would ensure consistent behavior across all platforms.

You're right bot, but I prefer minimizing the impact of this change for now. If 3.11.11 fails on Linux, it won't be worse than the current absence of a build. But if 3.11.11 breaks something on Mac or Windows, we'll have to fix it right away when we could have just stayed on a known working version.

dl1com (approved):

PR #748, opened 2025-01-23
Update ayab-translation-master.tsv for chinese translations

junelee220:
i try to make chinese translations but the only place I'm not too sure is"pass", in the "KnitProgress" section.


dl1com:
Thanks a lot for your contribution!
We'll check in the next dev meeting if we can incorporate the change in the 1.0 release.

dl1com:
Is there any chance to resolve the conflict in the tsv automatically? Otherwise, I'll write a script to do so... :D

dl1com (approved):

Issue #769, opened 2025-03-16
[BUG] Broken audio on macOS since the switch to QSoundEffect


jonathanperret:
**Environment**
AYAB software version: `main` commit d30d1ec40877e4d22518f39e6cfd687cfe9c9c15 (basically 1.0.0 plus #730)
Computer/OS: macOS 15.3.1 on MacBook Pro M1

**Describe the bug**

* Sounds emitted by the AYAB desktop app are cut off, late or entirely missing.
* Changing the OS's default audio output (e.g. switching from speakers to headphones) has no immediate effect on the AYAB app.

**To Reproduce**
1. Launch AYAB
2. Load any image
3. Select Simulation as the serial port
4. Enable Infinite Repeat
5. Click Knit
6. Listen to the "ding" sounds that should be emitted about once per second. On the built-in speakers, they get slightly cut-off as they overlap. On my AirPods, it's worse and many glitches can be heard. Also, sound often gets interrupted when resizing the AYAB window for example.
7. Switch to another audio output using the icon in the top menu bar: AYAB keeps emitting sounds on the previously selected device.

**Expected behavior**

AYAB release 1.0.0 does not exhibit any of the behaviors described above.

jonathanperret:
Fixed by #772.

PR #771, opened 2025-03-19
Upgrade `simpleaudio` to an updated fork that supports Python 3.12

### LINKED ISSUE #725: Hard crash with Python 3.12 and simpleaudio
**Environment**
AYAB software version: main / 1.0.0-beta3 / 02d2c5d0f57b138d9f0869aecc40feaa30642ee2
Computer/OS: Ubuntu LTS 20 + Python **3.12** (not 3.11 as recommended)
Knitting machine: simulator
AYAB hardware: simulator

**Describe the bug**
After starting the knitting simulator the first three lines have been 'knitted' by the simulator.
Every time after the third row is done, the application crashed hard.
No exit code, no error log, no noting.
Even when running it from vscode with the debugger enabled, it crashed without any hint.

**To Reproduce**
1. Use latest ayab-desktop (1.0.0-beta3) with python **3.12** an
2. Load an image
3. Select the simulator
4. Start knitting

**Expected behavior**
The application simulates the knitting process. No crash expected.

**Additional context**
I am using pythoon 3.12 instead of the recommended version 3.12. Using 3.11 would have been much more work, as it is not the default python version that comes with Ubuntu 24.04. (and I am not so familiar with using differen python environments in parallel)
However I could understand that this is not a "real" bug, if I am not using the recommended pyhton version.
But for future releases I wanted to point to this problem, as it might help another developer who works on the next version.

I was able to point the Problem down to this issue: https://github.com/hamiltron/py-simple-audio/issues/72
The described fix (`pip install -U --force git+https://github.com/cexen/py-simple-audio.git`) worked for me.

But as already mentioned. I am by far no expert in python or the implications of using different `PyMem_*` functions. 
(I could try to create a pull request if this helps)

So maybe one of the maintainers has a better idea for the future?
(and if it is only to write a debug message and hint the user to mute sounds in the settings page)

jonathanperret:
## Problem

Fixes #725 (without breaking macOS audio, see #769).

## Proposed solution

We use a fork named `simpleaudio-patched` that is compatible with recent Python versions and has up-to-date binary builds.

## 🔬 Testing

Here's a test build that should be used to check audio works properly on all supported platforms: https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.1-simpleaudio2

I've tested that this release has correct audio in macOS 15.3.1.

## Summary by CodeRabbit

- **Chores**
	- Enhanced the macOS build process to ensure robust support for audio playback.
	- Updated the audio dependency to a patched version, providing improved performance and compatibility across macOS architectures.

### DIFF
--- a/.github/workflows/build-multi-os.yml
+++ b/.github/workflows/build-multi-os.yml
@@ -351,6 +351,10 @@ jobs:
           python -m pip download --only-binary=:all: --platform macosx_11_0_arm64 "$(grep Pillow requirements.build.txt)"
           delocate-fuse Pillow*arm* Pillow*x86* -w tmp-wheel/
 
+          python -m pip download --only-binary=:all: --platform macosx_10_9_x86_64 "$(grep simpleaudio requirements.build.txt)"
+          python -m pip download --only-binary=:all: --platform macosx_11_0_arm64 "$(grep simpleaudio requirements.build.txt)"
+          delocate-fuse simpleaudio*arm* simpleaudio*x86* -w tmp-wheel/
+
           python -m pip download --only-binary=:all: --platform=macosx_10_9_universal2 "$(grep bitarray requirements.build.txt)"
 
           python -m pip install tmp-wheel/*.whl bitarray*.whl
--- a/requirements.build.txt
+++ b/requirements.build.txt
@@ -6,7 +6,7 @@ Pillow==10
 pyserial==3.5
 sliplib==0.6.2
 bitarray==2.9.2
-simpleaudio==1.0.4
+simpleaudio-patched==1.0.5
 numpy==1.26.4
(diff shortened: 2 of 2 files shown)


dl1com:
I reverted PR #730 so we got a clean main branch again for now, so removal of the Qt sound code can be removed from this PR

jonathanperret:
> I reverted PR #730 so we got a clean main branch again for now, so removal of the Qt sound code can be removed from this PR

Right that's cleaner. I've rebased this PR and created a new test build.

VIPQualityPost:
Does this need testing on real hardware?
On Ubuntu 24.10 with Simulator all of the audio is working for me (start chime + row bell)

dl1com (approved):

PR #773, opened 2025-03-19
Create pull_request_template.md

dl1com:


jonathanperret:
That's a nice initiative, thanks!
One first thing that comes to mind when reading the proposed template is that instructions for the PR author could be moved into HTML comments like we did for [issue templates](https://github.com/AllYarnsAreBeautiful/ayab-desktop/commit/79fc553d5eb969c2215f8c16142539e3ff00a1e2).

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Type of change
+
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] This change requires a documentation update
+- [ ] This change requires a translation update
```
jonathanperret, line 22: I'm not convinced a list like this is very practical — I would drop the "Type of change" section altogether unless we actually have a different process for different types of change. For example, "This change requires a documentation update" is useful in theory but unless someone painstakingly goes through PRs looking for those that need a doc update, the checkbox ends up being little more than an administrative annoyance. Just my opinion, of course.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
...
+
+# How Has This Been Tested?
```
jonathanperret, line 24: I would reword this into "How can this be tested?" or just "How to test" — this would put the focus on more testing being done by reviewers/testers rather than relying too much on the author having covered all bases.

In my opinion it's fine to ask for help if you can't thoroughly test a change you're proposing.

Issue #776, opened 2025-03-27
[FEATURE] Store Memo information in PNG file metadata 


t0mpr1c3:
Goal:
* For feature parity with Brother hardware/firmware, AYAB should implement storage and display of Memo information.

Problem:
* Currently the PNG input format used by AYAB has no way to store Memo information. 
* ~Memos are not displayed to the user (https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/250).~
* ~Memo information is lost when converting from patterns stored as DAK files.~

Suggested solution:
* Encode Memo data in the `Comment` field of the PNG file metadata. I suggest one character per line. A single digit 0-9 would suffice for most purposes, _and is best for easy editing_ ~but RFC 4648 encoding in base 64 is more performant~. 
* ~Change the `Software` field of the PNG file metadata to "AYAB". Memo information will only be extracted from the `Comment` field after testing that the `Software` keyword is "AYAB".~ _The comment field should have an initial header like "AYAB:"_ 
* ~Display Memo information to the user in the Knitting Progress window (lower window in UI)~
* ~Reverse engineer DAK file formats to extract Memo data.~

Alternative solution:
* Memo data could also be encoded using an additional pixel at the beginning or end of each row, which is not displayed as part of the pattern. This solution is more complex to implement and less satisfactory because the PNG will no longer display correctly as a graphical image.

Comment:
* Encoding Memo information in the text metadata would not preclude later changes to the colour/transparency encoding for additional features such as line endings, pixels that are not to be knitted, and so on.
* See also #250, #777

t0mpr1c3:
Note: We also need a code meaning "no memo information" but 0 works, I think

dl1com:
To throw another possible implementation in the ring:, just as an idea:
using a human readable text file with the same name as the PNG.
How the information is stored in this file is TBD (e.g. each line in the text file relates to the corresponding line in the PNG. Actions are encoded in letters/symbols).

Pros:

- no need to change the PNG
- no "hidden Infos" in the PNG
- no separate tooling required to edit and store memo data
- memo info is human read- and editable in any text editor
- memo file can be tracked in source control, diffs possible


Variant:
Put these files together in a zip file, enable ayab to open zip files 

- .zip ending = obvious indication that the pattern includes memos
- at least windows has a transparent handling of zip files in the file explorer

jonathanperret:
[...]

PR #779, opened 2025-03-29
Extract memo information from PNG file Comment tag and display in Knitting Progress window

t0mpr1c3:
Addressing #776

### DIFF
--- a/src/main/python/main/ayab/knitprogress.py
+++ b/src/main/python/main/ayab/knitprogress.py
@@ -36,6 +36,8 @@ if TYPE_CHECKING:
     from .engine.status import Status
 
 
+# TODO (TP): display memo information
+
 class KnitProgress(QTableWidget):
     """
     Class for the knit progress window, implemented as a subclass of `QScrollArea`.
@@ -44,8 +46,10 @@ class KnitProgress(QTableWidget):
     @date   June 2020
     """
 
-    green = 0xBBCCBB
-    orange = 0xEECC99
+    green = 0x99DD99
+    orange = 0xEECC77
+    grey = 0xCCCCCC
+    smoke = 0xF5F5F5
 
     def __init__(self, parent: GuiMain):
         super().__init__(parent.ui.graphics_splitter)
@@ -87,6 +91,7 @@ class KnitProgress(QTableWidget):
 
         if (
             status.line_number != self.previousStatus.line_number
+            or status.memo != self.previousStatus.memo
             or status.current_row != self.previousStatus.current_row
             or status.color_symbol != self.previousStatus.color_symbol
             or status.carriage_type != self.previousStatus.carriage_type
@@ -124,8 +129,6 @@ class KnitProgress(QTableWidget):
         if self.columnCount() != len(columns):
             self.setColumnCount(len(columns))
         n_cols = len(columns)
-        if n_cols < 4:
-            self.hideColumn(5)
         self.instantiate_row_from_columns(status, columns)
 
         self.previousStatus = status
@@ -136,6 +139,7 @@ class KnitProgress(QTableWidget):
     def load_columns_from_status(
         self, status: Status, columns: List[QTableWidgetItem]
     ) -> None:
+        columns.append(self.__memo(status.memo))
         for c in range(0, len(status.bits)):
             needle = status.knit_start_needle + c
             needle_number_from_r1 = needle - status.machine_width // 2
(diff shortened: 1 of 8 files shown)


jpcornil-git:
Why not using a JSON string to store memo ?
This would allow to:
- add more data moving forward, e;g. version, carriage to use for a row, colors, ...
- not be limited to single digit number
- just use a json.loads to get a dictionnary with everything in one line of code rather than parse string

Why not use an new exif or native iTXt "ayab" key rather than use a "Comment" key with value starting wth "AYAB:" ?

See https://discord.com/channels/1079442610332110849/1094306201451384944/1355294497696911410

t0mpr1c3:
> Why not using a JSON string to store memo ? This would allow to:
> 
>     * add more data moving forward, e;g. version, carriage to use for a row, colors, ...
> 
>     * not be limited to single digit number
> 
>     * just use a json.loads to get a dictionnary with everything in one line of code rather than parse string
> 
> 
> Why not use an new exif or native iTXt "ayab" key rather than use a "Comment" key with value starting wth "AYAB:" ?
> 
> See https://discord.com/channels/1079442610332110849/1094306201451384944/1355294497696911410

There is no need from the point of view of feature parity, it is harder to implement, it makes manual editing more difficult.

jpcornil-git:
Not sure I understand why you said it is harder to implement (on the contrary) and that it makes manual editing more difficult (same to me), see below

Current code
info["Comment"] = 'AYAB:12345'
``` python
            if "Comment" in self.image.info and len(str(self.image.info["Comment"])) > 0:
                comment = str(self.image.info["Comment"])
                if comment.startswith("AYAB:"):
                    # update memo information
                    self.memos = []
                    for i in range(len(comment) - 5):
                        self.memos.append(int(comment[i + 5]))
```

Use JSON instead of raw/single digit data

info["Comment"] = 'AYAB:{"memo":[1,12,23,34,45], "color":[1,2,3,4,5], "carriage":["K", "K", "L", "L", "G"], "version":1}'

``` python
            if "Comment" in self.image.info and len(str(self.image.info["Comment"])) > 0:
                comment = str(self.image.info["Comment"])
                if comment.startswith("AYAB:"):
                    metadata = json.loads(comment[5:])
                    # update memo information
                    self.memos = metadata["memo"]
```
=> no parsing in the code anymore, extend beyond "memo"

Use JSON with a dedicated metadata key

info["AYAB"] = '{"memo":[1,12,23,34,45], "color":[1,2,3,4,5], "carriage":["K", "K", "L", "L", "G"], "version":1}'

``` python
            if "AYAB" in self.image.info:
                metadata = json.loads(self.image.info["AYAB"])
                # update memo information
                self.memos = metadata["memo"]
```

=> Same as above + no need to check for a string prefix

Code is shorter/generic and give you much more flexibility for free ?

t0mpr1c3:
The format implemented in the code is consistent with https://github.com/AllYarnsAreBeautiful/ayab-patterns/pull/7

If you want to change one, you'll want to change the other.

jonathanperret:
I've added a commit that points the patterns submodule to include the already-enhanced patterns (up to number 124 in `kh930`).

I have a couple of nits to pick with the code as it is (mainly the issues reported by SonarQube above) but the UI part feels ready for initial testing.

So I've triggered a build from this PR so that we can collect tester feedback: https://github.com/jonathanperret/ayab-desktop/releases/tag/1.0.0-memo1

src/main/python/main/ayab/knitprogress.py
```diff
@@ -44,8 +46,10 @@ class KnitProgress(QTableWidget):
     @date   June 2020
     """
 
-    green = 0xBBCCBB
-    orange = 0xEECC99
+    green = 0x99DD99
+    orange = 0xEECC77
```
jonathanperret, line 50: It doesn't look like this change in colors is related to the topic of this PR, is it?
In any case, changing the color breaks the selection display for right-side stitches, it now displays `stitch Left-x`:
image

t0mpr1c3: Does changing the colour back fix that?

t0mpr1c3: there is some fragile code on line 293 of knitprogress.py

t0mpr1c3: fixed in https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/779/commits/274b3529bd2defd340039544e37eb5c77a1deb08

t0mpr1c3:
Invert and HFlip should be safe for Memos.

Repeat repeats Memos by the vertical scale factor.

Everything else is potentially unsafe and results in deletion of Memos.

t0mpr1c3:
I don't know how to freeze the Memo column. @X-sam ?

dl1com:
What's the status of this PR?
There are still some CodeRabbit and SonarQube findings to clarify.

t0mpr1c3:
Not impressed with the remaining CodeRabbit suggestions.

dl1com:
> Not impressed with the remaining CodeRabbit suggestions.

Please comment and resolve them, so we can proceed with the testing step.

But out of curiosity: https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/779#discussion_r2051707253 suggests improvements for handling the transforms where memo is about to be deleted, something you also mentioned in https://github.com/AllYarnsAreBeautiful/ayab-desktop/pull/779#issuecomment-2767645414. So may be worth to follow the suggestion?

src/main/python/main/ayab/knitprogress.py
```diff
@@ -124,8 +132,6 @@ def update_progress(self, status: Status) -> None:
         if self.columnCount() != len(columns):
             self.setColumnCount(len(columns))
         n_cols = len(columns)
-        if n_cols < 4:
-            self.hideColumn(5)
```
jonathanperret, line 128: Removing those two lines (which do look like dead code, as the number of columns is synchronized with the size of the `columns` array earlier, so column 5 can't exist if `len(columns) < 4`) makes the `n_cols` assignment above obsolete as well.

PR #800, opened 2025-10-04
Change default log level to INFO

jpcornil-git:
Remove default logging.basicConfig in Communication class init; it has no effect and should be set once in main.py anyway.

The DEBUG log level should not be set in the root log as it may be used as the default for other modules and thus unnecessarily increase verbosity.

I added a setlevel(DEBUG) to the Communication's logger to keep the intended DEBUG level there but this should probably be set to INFO as well (or removed, added when required/while debugging).

## Summary by CodeRabbit

- Refactor
  - Switched to per-instance logging to isolate component logs and avoid altering global logging behavior.
- Chores
  - Lowered default file log level from DEBUG to INFO to reduce noisy entries while keeping formats and timestamps, yielding cleaner logs for routine monitoring.

### DIFF
--- a/src/main/python/main/ayab/engine/communication.py
+++ b/src/main/python/main/ayab/engine/communication.py
@@ -72,8 +72,8 @@ class Communication(object):
     def __init__(self, serial: Optional[serial.Serial | WebsocketSerial] = None):
         """Create an AyabCommunication object,
         with an optional serial communication object."""
-        logging.basicConfig(level=logging.DEBUG)
         self.logger = logging.getLogger(type(self).__name__)
+        self.logger.setLevel(logging.DEBUG)
         self.__ser = serial
         self.__driver = sliplib.Driver()
         self.rx_msg_list: list[bytes] = list()
--- a/src/main/python/main/main.py
+++ b/src/main/python/main/main.py
@@ -53,7 +53,7 @@ class AppContext(ApplicationContext):  # type: ignore # 1. Subclass ApplicationC
         logfile = path.join(self.userdata_path, "ayab_log.txt")
         logging.basicConfig(
             filename=logfile,
-            level=logging.DEBUG,
+            level=logging.INFO,
             format="%(asctime)s %(name)-8s %(levelname)-8s %(message)s",
             datefmt="%y-%m-%d %H:%M:%S",
         )


jonathanperret:
Good catch on the extra `basicConfig`, I agree it does not make sense (and per the python docs it does nothing except maybe cause a race condition in some cases).

In fact storing a logger in `self.logger` is explicitly discouraged in the docs: https://docs.python.org/3/howto/logging-cookbook.html#using-loggers-as-attributes-in-a-class-or-passing-them-as-parameters — it would be better to move to the pythonic approach of a module-level logger, though the wrong pattern appears to be used multiple times in the current code so that would be a wider change.

In any case, configuring the logger in the `Communication` constructor is not the right place to do it because it will override any configuration that may have been set up in the proper place (i.e. in `main` or a configuration file), and reset it on every instance creation (since loggers are singletons, `logging.getlogger` will always return the same instance for a given name).

A better alternative would be to add a call to `logging.getLogger("…Communication").setLevel(DEBUG)` in `main.py` right after the `basicConfig` call). A generalized version of this would be to call `dictConfig` and configure the desired level for any logger of interest.

Regarding the default logging level, I agree that `DEBUG` for everything by default probably isn’t right. However, until there is a user-accessible way to enable increased logging I’d rather err on the side of logging too much, in the interest of being able to diagnose issues from the field. 

Keeping just the `Communication` logger at `DEBUG` and moving everything else back to `INFO` may well keep sufficient logging, but I’d want to look at which messages we’d be removing from the logs by doing so to assess whether it is OK to make the change.

jpcornil-git:
You could do something like illustrated below if you want to conditionally enable DEBUG level without a UI/preference file setting, in main.py:
``` python
...
        logging.basicConfig(
            filename=logfile,
            level=logging.DEBUG if os.getenv('AYAB_DEBUG') else logging.INFO,
            format="%(asctime)s %(name)-8s %(levelname)-8s %(message)s",
            datefmt="%y-%m-%d %H:%M:%S",
        )
        console = logging.StreamHandler()
...
```
=> loglevel is INFO unless you define AYAB_DEBUG in your environment

List of logging/logger.debug calls with suggested loglevel modification (first column)

``` bash
$ grep -r 'logging.debug' src/*
  src/main/python/main/ayab/ayab.py:        logging.debug("Quitting")
  src/main/python/main/ayab/utils.py:    logging.debug(f"MessageBox {message_type}: '{message}'")
  src/main/python/main/ayab/gui_fsm.py:        self.NO_IMAGE.entered.connect(lambda: logging.debug("Entered state NO_IMAGE"))
  src/main/python/main/ayab/gui_fsm.py:            lambda: logging.debug("Entered state TESTING_NO_IMAGE")
  src/main/python/main/ayab/gui_fsm.py:            lambda: logging.debug("Entered state CONFIGURING")
  src/main/python/main/ayab/gui_fsm.py:        self.CHECKING.entered.connect(lambda: logging.debug("Entered state CHECKING"))
  src/main/python/main/ayab/gui_fsm.py:        self.KNITTING.entered.connect(lambda: logging.debug("Entered state KNITTING"))
  src/main/python/main/ayab/gui_fsm.py:        self.TESTING.entered.connect(lambda: logging.debug("Entered state TESTING"))


$ grep -r 'logger.debug' src/*
  src/main/python/main/ayab/firmware_flash.py:        self.__logger.debug("port " + str(self.port))
  src/main/python/main/ayab/firmware_flash.py:        self.__logger.debug(exec_command)
  src/main/python/main/ayab/version_checker.py:        self.logger.debug(
  src/main/python/main/ayab/version_checker.py:        self.logger.debug("Getting %s", latest_relase_url)
  src/main/python/main/ayab/version_checker.py:            self.logger.debug(
  src/main/python/main/ayab/version_checker.py:                self.logger.debug("Latest version is %s at %s", latest_version, url)
  src/main/python/main/ayab/version_checker.py:            self.logger.debug("Cleaning up")
W src/main/python/main/ayab/engine/communication.py:        self.logger.debug("unknown message: ")  # drop crlf
  src/main/python/main/ayab/engine/control.py:        self.logger.debug(msg)
  src/main/python/main/ayab/engine/control.py:        self.logger.debug("sending blank line as final line=%d", requested_line)
  src/main/python/main/ayab/engine/mode.py:            control.logger.debug("COLOR " + str(color))
I src/main/python/main/ayab/engine/engine_fsm.py:        control.logger.debug("State CONNECT")
  src/main/python/main/ayab/engine/engine_fsm.py:        control.logger.debug("Port name: " + control.portname)
I src/main/python/main/ayab/engine/engine_fsm.py:        control.logger.debug("State VERSION_CHECK")
I src/main/python/main/ayab/engine/engine_fsm.py:                control.logger.debug("State INIT")
I src/main/python/main/ayab/engine/engine_fsm.py:                    control.logger.debug("State REQUEST_TEST")
I src/main/python/main/ayab/engine/engine_fsm.py:                    control.logger.debug("State REQUEST_START")
I src/main/python/main/ayab/engine/engine_fsm.py:                control.logger.debug("State CONFIRM_START")
W src/main/python/main/ayab/engine/engine_fsm.py:                control.logger.debug(
I src/main/python/main/ayab/engine/engine_fsm.py:                control.logger.debug("State RUN_KNIT")
I src/main/python/main/ayab/engine/engine_fsm.py:        control.logger.debug("State CONFIRM_TEST")
I src/main/python/main/ayab/engine/engine_fsm.py:                control.logger.debug("State RUN_TEST")
I src/main/python/main/ayab/engine/engine_fsm.py:            control.logger.debug("State DISCONNECT")
I src/main/python/main/ayab/engine/engine_fsm.py:        control.logger.debug("State FINISHED")
  src/main/python/main/ayab/engine/engine.py:        self.__logger.debug(self.config.as_dict())
```

PR #801, opened 2025-10-04
Adapt retry logic for communication with firmware

jpcornil-git:
This is to avoid latency issues with Wifi connections where delay is often more than 100m, e.g. from discord discussions

```
gjh@fedora:~$ ping ayab.local
PING ayab.local (192.168.0.129) 56(84) bytes of data.
64 bytes from 192.168.0.129: icmp_seq=1 ttl=64 time=344 ms
64 bytes from 192.168.0.129: icmp_seq=2 ttl=64 time=558 ms
64 bytes from 192.168.0.129: icmp_seq=3 ttl=64 time=181 ms
64 bytes from 192.168.0.129: icmp_seq=4 ttl=64 time=400 ms
64 bytes from 192.168.0.129: icmp_seq=5 ttl=64 time=632 ms
64 bytes from 192.168.0.129: icmp_seq=6 ttl=64 time=241 ms
64 bytes from 192.168.0.129: icmp_seq=7 ttl=64 time=468 ms
64 bytes from 192.168.0.129: icmp_seq=8 ttl=64 time=86.6 ms
64 bytes from 192.168.0.129: icmp_seq=9 ttl=64 time=523 ms
64 bytes from 192.168.0.129: icmp_seq=10 ttl=64 time=546 ms
64 bytes from 192.168.0.129: icmp_seq=11 ttl=64 time=149 ms
64 bytes from 192.168.0.129: icmp_seq=12 ttl=64 time=377 ms
64 bytes from 192.168.0.129: icmp_seq=13 ttl=64 time=617 ms
64 bytes from 192.168.0.129: icmp_seq=14 ttl=64 time=217 ms
64 bytes from 192.168.0.129: icmp_seq=15 ttl=64 time=455 ms
64 bytes from 192.168.0.129: icmp_seq=16 ttl=64 time=58.7 ms
64 bytes from 192.168.0.129: icmp_seq=17 ttl=64 time=286 ms
64 bytes from 192.168.0.129: icmp_seq=18 ttl=64 time=515 ms
64 bytes from 192.168.0.129: icmp_seq=19 ttl=64 time=126 ms
64 bytes from 192.168.0.129: icmp_seq=20 ttl=64 time=364 ms
64 bytes from 192.168.0.129: icmp_seq=21 ttl=64 time=792 ms
64 bytes from 192.168.0.129: icmp_seq=22 ttl=64 time=195 ms
```

## Summary by CodeRabbit

* **Refactor**
  * Initialization now starts immediately after a successful API6 token validation.
  * Removed automatic scheduling of a retry during API6 initialization; the flow now enters "initializing firmware" without queuing a retry.
  * Increased the default retry interval from 0.1s to 1.0s (users may notice slower, less frequent automatic retries).

### DIFF
--- a/src/main/python/main/ayab/engine/engine_fsm.py
+++ b/src/main/python/main/ayab/engine/engine_fsm.py
@@ -121,6 +121,7 @@ class StateMachine(QStateMachine):
                 control.api_version = param
                 control.state = State.INIT
                 control.logger.debug("State INIT")
+                control.com.req_init_API6(control.machine)
                 return Output.NONE
             else:
                 control.logger.error(
@@ -153,7 +154,6 @@ class StateMachine(QStateMachine):
                 control.logger.error("Error initializing firmware: " + str(param))
                 return Output.ERROR_INITIALIZING_FIRMWARE
         # else
-        StateMachine.retry(control.com.req_init_API6, (control.machine,))
         return Output.INITIALIZING_FIRMWARE
 
     @staticmethod


jonathanperret:
I'm afraid bumping it the retry interval to 1 second could perceivably slow down the initialization time on USB. If a longer delay is necessary for WebSockets, could it be made transport-dependent?

Also: as far as I know, the retry logic was only introduced because of the ~1 second delay imposed by the Arduino bootloader upon reset, which causes the MCU to be unresponsive since the firmware hasn't been entered yet, combined with the automatic reset on serial port open.

USB (and WebSocket) are already reliable protocols (both will resend any lost/corrupted packets) so beyond that initial period where the firmware is absent, there should not be any need to resend instructions from `ayab-desktop`.

In particular, the logic to [resend the `init` request](https://github.com/AllYarnsAreBeautiful/ayab-desktop/blob/6a8a864a491719633d3eb0fdff036d35027ff824/src/main/python/main/ayab/engine/engine_fsm.py#L156) seems entirely unnecessary, since we only send that message once we've received a response to the previous `info` request, and at this point the firmware is known to be operational.

That logic is also potentially harmful since it can cause the firmware to receive multiple `init` requests, the second and further ones generating an "invalid state" error since the firmware may have transitioned to another state in the interval, as evidenced in  the logs you analyzed that showed a `confirmInit MachineInvalidState(0xef)` response coming after a successful `confirmInit` response due to processing of extra `init` requests.

I would therefore suggest:
 - keeping the current relatively short retry interval of `info` requests, helping to initiate communication as soon as the firmware is available (a few extra requests potentially queuing up on Wi-Fi will be harmless as the requests and responses are small and idempotent);
 - removing the retry logic from the `init` call, letting a single request take the time it needs to reach the firmware. If that one request should not reach the firmware, there is no point in continuing further because it would mean the communication is unreliable or the firmware has crashed, which are both unrecoverable conditions.

jpcornil-git:
Agree with you that a retry is a priori never required except for devices that are reset when the connection is established and only until there.

=> why not move the `control.com.req_init_API6 ` call in `_API6_version_check` (single call before transitioning to the `init` state only waiting for an acknowledge)

``` python
...
    def _API6_version_check(control: Control, operation: Operation) -> Output:
        token, param = control.check_serial_API6()
        if token == Token.cnfInfo:
            if param >= control.FIRST_SUPPORTED_API_VERSION:
                control.api_version = param
                control.state = State.INIT
                control.logger.debug("State INIT")
                control.com.req_init_API6(control.machine)
                return Output.NONE
...
```

jonathanperret:
> Agree with you that a retry is a priori never required except for devices that are reset when the connection is established and only until there.
> 
> => why not move the `control.com.req_init_API6 ` call in `_API6_version_check` (single call before transitioning to the `init` state only waiting for an acknowledge)
> 
> ```python
> ...
>     def _API6_version_check(control: Control, operation: Operation) -> Output:
>         token, param = control.check_serial_API6()
>         if token == Token.cnfInfo:
>             if param >= control.FIRST_SUPPORTED_API_VERSION:
>                 control.api_version = param
>                 control.state = State.INIT
>                 control.logger.debug("State INIT")
>                 control.com.req_init_API6(control.machine)
>                 return Output.NONE
> ...
> ```

Yes, this is exactly what I had in mind. This should prevent the double-init failure.

jonathanperret (approved):
Looks good to me, let's get some feedback from testing on actual hardware.

jonathanperret:
Fix was reported to work on Discord, I'm merging this. Thanks @jpcornil-git!

Issue #726, opened 2024-11-11
[FEATURE] Rename 'Infinite Repeat' to ' Vertical Repeat'


jllyh:
This is more of a usability / clarity issue.

Can "Infinite Repeat" be renamed "Vertical Repeat"?  "Infinite" doesn't imply direction. I had assumed that it meant a horizontal repeat, and it isn't clarified anywhere in the user instructions. 
Also, if it's renamed "Vertical Repeat," then "Horizontal Repeat" will be the logical choice when / if that is added.

jonathanperret:
"Infinite Vertical Repeat" sounds like a good option.

jonathanperret:
Let's remember to change it in the Preferences panel as well.

PR #806, opened 2026-04-21
fix(i18n): add 'vertical' to inifinite vertical repeat strings

raemond:
Addresses #726 

I've updated the strings that refer to "Infinite Repeat" to "Infinite Vertical Repeat" across all translations in both the Settings and Preferences.

I updated the internal lookup key, just in case we introduce other forms of infinite repeat (e.g. horizontal) in future and need to differentiate.

## Summary by CodeRabbit

* **Style**
  * Updated the checkbox label in the options tab from "Infinite Repeat" to "Infinite Vertical Repeat" for clearer description of the repeat functionality direction.


X-sam (commented):
Why change the key name? This change would be much safer and have no code changes if we simply updated the translations.

raemond:
My understanding of the code is that it uses the en_US string as (part of) the lookup key. If we want to change the en_US string, the key would change implicitly. Ideally, we'd use non-language specific keys in the translation file as references to avoid this but that felt like a bigger task and possibly overkill for changing one string.

PR #809, opened 2026-09-10
Add Czech (cs_CZ) translation

infanap:
Adds a `cs_CZ` column with all 122 strings to `ayab-translation-master.tsv`.

- The `.ts`/`.qm` files are generated by `ayab_trans.pl` at build time and the language list in Preferences is built from the available `.qm` files, so no code changes are needed; the language shows up as "čeština".
- Terminology follows the Czech knitting-machine manuals (carriage = saně, single/double bed = jednolůžkové/dvoulůžkové, circular = hadicové pletení).
- All other columns are unchanged (123 lines changed only by the appended column).

Discussed with Victoria by e-mail; happy to adjust any string. Czech versions of the website and the manual can follow once there is a place for them.


infanap:
Thanks for approving the workflow runs, @jonathanperret. I had a look at the two red jobs; neither is caused by this PR:

- **Linux**: `wget` of `python3.11.13-cp311-cp311-manylinux_2_28_x86_64.AppImage` from niess/python-appimage returns 404. The `python3.11` release there now only carries `python3.11.16-...`, so the pinned file name needs a bump. The same step has been failing on other branches since February (including the 1.1.0-rc1 push on 6 Sep).
- **Windows**: `fbs freeze` stops with `FileNotFoundError: Could not find msvcr110.dll on your PATH` (fbs `_add_missing_dlls`). It looks like the `windows-2022` runner image no longer ships the VC++ 2012 redistributable; the Windows job was still green on 6 Sep.

The step "Convert UI and translation files" (which runs `ayab_trans.pl` on the new `cs_CZ` column) passes on all three platforms, and the macOS build is green. Happy to adjust any string, or to rebase once the CI is fixed.

jonathanperret:
> Thanks for approving the workflow runs, @jonathanperret. I had a look at the two red jobs; neither is caused by this PR:
> 
> * **Linux**: `wget` of `python3.11.13-cp311-cp311-manylinux_2_28_x86_64.AppImage` from niess/python-appimage returns 404. The `python3.11` release there now only carries `python3.11.16-...`, so the pinned file name needs a bump. The same step has been failing on other branches since February (including the 1.1.0-rc1 push on 6 Sep).
> * **Windows**: `fbs freeze` stops with `FileNotFoundError: Could not find msvcr110.dll on your PATH` (fbs `_add_missing_dlls`). It looks like the `windows-2022` runner image no longer ships the VC++ 2012 redistributable; the Windows job was still green on 6 Sep.

Thanks for looking into the build failures. I agree they have nothing to do with your PR, which looks OK to me.

The `python-appimage` one is a known recurring issue (see https://github.com/AllYarnsAreBeautiful/ayab-desktop/issues/721), I'm working on fixing it again.

The Windows one is new and indeed very recent, probably the result of some runner image update, I need to look into it.

